PR #3, opened 2020-01-30
Updates/mos 91 rename grid internal

svjoshua:


svjoshua:
will re-open with Jira friendly name

PR #20, opened 2020-02-11
MOS-48 Sanitize and expand example data

svjoshua:


svjoshua:
Will re-open data changes only for easier review

PR #27, opened 2020-02-14
Updates validation schema for additional actions

svjoshua:


owenallenaz:
No longer valid with changes done a few sprints back.

PR #51, opened 2022-03-31
MOS-621: Document batch 3 of components

JAndresHJGlobant:
## What's included?
Documentation of the input settings and all extra information needed (i.e. ways of invoking it inside a form) inside the Form.stories.mdx file for the following components:

- FormFieldRadio
- FormFieldTable
- FormFieldText
- FormFieldTextArea
- FormFieldTextEditor
- FormFieldToggleSwitch


src/Forms/Form/Form.stories.mdx
```diff
@@ -56,11 +56,313 @@ const fields = useMemo(
 	[]
 );
 ```
+### FormFieldRadioButton
+- A group of radio buttons allows users to select a single item from a list of possible options.
+- All possible options are exposed up front for comparison.
+- Users can only make a single selection from the list of possible options.
+- **Playground**: https://simpleviewinc.github.io/sv-mosaic/?path=/story/formfields-formfieldradio--playground
```
mauriciodeleonc, line 63: I think it's better if we keep the word Playground as link. You could change it to [**Playground**](/?path=/story/formfields-formfieldradio--playground) and the word Playground will become the link. Btw if you write the link like this it will go to the address without caring if it's kube, simpleview.github.com, etc.

src/Forms/Form/Form.stories.mdx
```diff
@@ -56,11 +56,313 @@ const fields = useMemo(
 	[]
 );
 ```
+### FormFieldRadioButton
```
mauriciodeleonc, line 59: Let's take 1 level up all headings, for example this would be ## FormFieldRadioButton, props would be ### Props, etc.

src/Forms/Form/Form.stories.mdx
```diff
@@ -56,11 +56,313 @@ const fields = useMemo(
...
+- A group of radio buttons allows users to select a single item from a list of possible options.
+- All possible options are exposed up front for comparison.
+- Users can only make a single selection from the list of possible options.
+- **Playground**: https://simpleviewinc.github.io/sv-mosaic/?path=/story/formfields-formfieldradio--playground
 
-	
+#### Props (Input Settings)
+* **options** - `array` of `object` optional - Predefined set of mutually exclusive options.
+	* **label** - `string` - Description of the option.
```
mauriciodeleonc, line 67: Label and value are missing the required tag

src/Forms/Form/Form.stories.mdx
```diff
@@ -56,11 +56,313 @@ const fields = useMemo(
...
+	* **value** - `string` - Value of the option, this is stored in the Form state.
+
+#### How to use in a form?
+```ts
+const fields = useMemo(
+		() =>
+			[
+				{
+					name: "radio",
```
mauriciodeleonc, line 76: Let's leave only the type and input settings (on all examples) so we don't over repeat code that doesn't belong to this specific field. All other generic props will be on their own section.

src/Forms/Form/Form.stories.mdx
```diff
@@ -56,11 +56,313 @@ const fields = useMemo(
...
+- **Playground**: https://simpleviewinc.github.io/sv-mosaic/?path=/story/formfields-formfieldtext--playground
+
+#### Props (Input Settings)
+
+* **htmlFor** - `string` optional - Specifies which form element a label is bound to.
+* **maxCharacters** - `number` optional - Defines the limit of characters that are allow to type in the input element.
+* **multiline** - `boolean` optional - When is enabled the text field will expand its height.
+* **placeholder** - `string` optional - The hint displayed in the input before the user enters a value.
+* **prefixElement** - `JSX.Element` optional - Icon at the beginning of the text field.
+* **size** - `enum { xs = '100px', sm = 280px, md = "450px", lg = "620px" }` required - The size of the component.
```
mauriciodeleonc, line 117: For the size please use the following text: "Size of the field. Could be either one of the predefined sizes (please see Sizes section), or a custom size. This property can be internally overwritten if the field is placed inside of a row with other fields." and the following type: "Size | string". There will be a Sizes section to explain more about them, also this way if sizes change we won't need to update it everywhere.

src/Forms/Form/Form.stories.mdx
```diff
@@ -56,11 +56,313 @@ const fields = useMemo(
...
+* **extraActions** - `array` of `object` optional - Possible actions that the table could execute and display.
+	* **label** - `string` - Describes the action.
+	* **icon** - `MosaicMIcon` - MUI icon.
+	* **actionFnc** - `actionFnc: (rowIndex: number) => void;` - Function that is executed when the corresponding icon is clicked.
+* **headers** - `string[]` required - Array of the table headers.
+
+#### How to use in a form?
+
+```ts
+const possibleTableRows = [
```
mauriciodeleonc, line 244: Let's keep the examples only on how to invoke the field inside a form, all of this functions and variables could be inferred when reading the props.

src/Forms/Form/Form.stories.mdx
```diff
@@ -56,11 +56,313 @@ const fields = useMemo(
...
+
+const fields = useMemo(
+	() =>
+		[
+			{
+				disabled: false,
+				helperText: 'Helper text',
+				instructionText: 'Instruction text',
+				inputSettings: {
+					handleAddElement: addTableRow,
```
mauriciodeleonc, line 300: If we change this to just handleAddElement: myAddFuncion, then by reading this and the prop list I think it's more than enough for devs to decide how the function should work like.

src/Forms/Form/Form.stories.mdx
```diff
@@ -56,11 +56,313 @@ const fields = useMemo(
...
+
+### FormFieldToggleSwitch
+- Toggle Switch allow users to switch between two possible states. They are commonly used to turn a specific setting on or off
+- Toggles should be used to turn on or off a preference, notification, or feature
+- Should be used when an instant response is required/desired
+- **Playground**: https://simpleviewinc.github.io/sv-mosaic/?path=/story/formfields-formfieldtoggleswitch--playground
+
+#### Props (Input Settings)
+
+* **toogleLabel** - `string` optional - This label is placed at the right of the switch.
```
mauriciodeleonc, line 339: Correction: toggleLabel*

mauriciodeleonc (changes requested):

src/Forms/Form/Form.stories.mdx
```diff
@@ -57,10 +87,538 @@ const fields = useMemo(
...
+		],
+	[]
+);
+```
+
+## FormFieldRadioButton
+- A group of radio buttons allows users to select a single item from a list of possible options.
+- All possible options are exposed up front for comparison.
+- Users can only make a single selection from the list of possible options.
+- **Playground**: https://simpleviewinc.github.io/sv-mosaic/?path=/story/formfields-formfieldradio--playground
```
mauriciodeleonc, line 336: This link is missing the update

src/Forms/Form/Form.stories.mdx
```diff
@@ -57,10 +87,538 @@ const fields = useMemo(
...
+```
+
+## FormFieldText
+- Text input fields are used to enter text information, numbers, e-mail addresses, or passwords. 
+- A text field consists of a label and a line with variations in width.
+- [**Playground**](/?path=/story/formfields-formfieldtext--playground)
+
+#### Props (Input Settings)
+
+* **htmlFor** - `string` optional - Specifies which form element a label is bound to.
```
mauriciodeleonc, line 384: I think htmlFor should be removed as inputSetting from all components. Since it should match the name then I think it's better if we just completely remove htmlFor and just "inject" the name value internally in the Col.tsx file

src/Forms/Form/Form.stories.mdx
```diff
@@ -57,10 +87,538 @@ const fields = useMemo(
...
+* **maxCharacters** - `number` optional - Defines the limit of characters that are allow to type in the input element.
+* **multiline** - `boolean` optional - When is enabled the text field will expand its height.
+* **placeholder** - `string` optional - The hint displayed in the input before the user enters a value.
+* **prefixElement** - `JSX.Element` optional - Icon at the beginning of the text field.
+* **size** - `Size | string` required - Size of the field. Could be either one of the predefined sizes (please see Sizes section), or a custom size. This property can be internally overwritten if the field is placed inside of a row with other fields.
```
mauriciodeleonc, line 389: Shouldn't size be optional? We already have a functionality that if no size gets passed then a default takes place, please check this.

mauriciodeleonc: Also I'm thinking we could remove size as an inputSettings prop and just inject it directly in Col.tsx (inject it from fieldDef), but this is something for another ticket.

src/Forms/Form/Form.stories.mdx
```diff
@@ -57,10 +87,538 @@ const fields = useMemo(
...
+);
+```
+
+## FormFieldTextArea
+- A text area is a multi-line plain-text editing control, is useful when you want to allow users to enter a sizeable amount of free-form text, but defining a specific height and width.
+- [**Playground**](/?path=/story/formfields-formfieldtextarea--playground)
+
+#### Props (Input Settings)
+
+* **htmlFor** - `string` optional - Specifies which form element a label is bound to.
```
mauriciodeleonc, line 424: Same comment from FormFieldText

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #60, opened 2022-04-06
MOS-618: Fix input settings from documentation

cristianvj23:
Fixed next issues in the Form documentation

- Take validators and types to the end of documentation.
- Remove “Props (Input settings)” subsection and replace as bullet point under description subsection.
- Rename “Default settings” to data


mauriciodeleonc (approved):
Approved but has minor non-breaking issues

PR #66, opened 2022-04-08
MOS-622: Change size use

cristianvj23:
# What's include:
- Changed way to import sizes. As developers who are creating forms they do not have to import Sizes, but just by writing the size, the form itself knows what the size is.


src/Forms/Form/Col.tsx
```diff
@@ -126,16 +126,18 @@ const Col = (props: ColPropsTypes) => {
 				const error = state?.errors[fieldProps.name] || "";
 
 				let maxSize: Sizes | string;
+				const SizeSelected = typeof type === "string" ? Sizes[currentField?.size] : currentField?.size;
```
mauriciodeleonc, line 129: I don't think this condition makes much sense, the prop "type" is used to know what type of component should be used, so it doesn't have connection with the size. I think you could only keep it as `const SizeSelected = Sizes[currentField?.size] ? Sizes[currentField?.size] : currentField?.size;`  (if the size passed by the developer exists on the Sizes enum then put that value, if not then pass the value passed by the dev)

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #79, opened 2022-05-03
MOS-640: Remove required from inputSettings

cristianvj23:
# What's include

- It was checked that inputSettings does not have the required property
- The verification was added in the Col component that required comes from currentField


mauriciodeleonc:
This PR will be closed as the ticket was created as a misunderstanding from my part.

PR #84, opened 2022-05-04
MOS-644: Make validators more robust

cristianvj23:
# What's include

Validations for:

- validateEmail
- validateNumber
- validateURL
- validateDateRange


src/forms/Form/validators.ts
```diff
@@ -13,14 +13,16 @@ export function validateEmail(str: string): string | undefined {
...
+		const isValidEmail = str
+			.toLowerCase()
+			.match(
+				/^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+			);
+	
+		if (isValidEmail) return;
+	}
 
 	return "The value is not a valid e-mail";
```
mauriciodeleonc, line 27: The error should only appear if the value is a string. The idea of this ticket is to only show error messages on the fields they belong, so for example, this checkbox component shouldn't show the error message
image

src/forms/Form/validators.ts
```diff
@@ -68,6 +70,8 @@ export function validateNumber(value: string): string | undefined {
 
 	if (!(!isNaN(Number(value)) && !isNaN(parseFloat(value)))) {
 		return "The value is not a number";
+	} else {
+		return;
```
mauriciodeleonc, line 74: Similar comment as above. You can test and make sure the solution complies with the ticket if you put a field that's not supposed to have that error message and doesn't show it.
image

src/forms/Form/validators.ts
```diff
@@ -80,18 +84,21 @@ export function validateNumber(value: string): string | undefined {
...
+		}
 	}
 
-	const pattern = new RegExp("^(https?:\\/\\/)?" + // protocol
-		"((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|" + // domain name
-		"((\\d{1,3}\\.){3}\\d{1,3}))" + // OR ip (v4) address
-		"(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*" + // port and path
-		"(\\?[;&a-z\\d%_.~+=-]*)?" + // query string
-		"(\\#[-a-z\\d_]*)?$", "i"); // fragment locator
+	return "The value is not a valid URL";
```
mauriciodeleonc, line 100: Similar comment as above
image

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #96, opened 2022-05-18
MOS-259: Specify default comparison

cristianvj23:
# What's include: 

- Added new prop default comparison to allow developr choise default filter comparison

### DIFF
--- a/src/components/internal/DataViewFilters.jsx
+++ b/src/components/internal/DataViewFilters.jsx
@@ -112,13 +112,14 @@ function DataViewFilters(props) {
 					{
 						active.map(filter => {
 							const Component = filter.component;
+							const filterData = props.filter[filter.name] || ((filter.comparisonDefault && filter.name === "title_with_comparisons") ? {comparison: filter.comparisonDefault} : {})
 							return (
 								<Component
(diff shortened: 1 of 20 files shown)


src/components/internal/DataViewFilters.jsx
```diff
@@ -112,14 +112,13 @@ function DataViewFilters(props) {
...
 							const Component = filter.component;
-							
 							return (
 								<Component
 									key={filter.name}
 									label={filter.label}
 									type={filter.type}
 									args={filter.args || {}}
-									data={props.filter[filter.name] || {}}
+									data={props.filter[filter.name] || (filter.comparisonDefault ? {comparison: filter.comparisonDefault} : {})}
```
mauriciodeleonc, line 121: This approach does solve the issue for the "Titles with comparisons" filter, but now breaks the "Categories with comparisons" because this change affects all. I suggest making the change directly on the component that renders "Titles with comparisons"; I believe it's DataViewFilterText.jsx

This is the error that appears.
image

Also remember to add some unit test to make sure the default option is the same one that renders when opening the component.

mauriciodeleonc (approved):

PR #97, opened 2022-05-19
MOS-660/updates-date-time-mui-pickers

JAndresHJGlobant:
## What is included?

- The update of the Time and Date picker following the migration guide https://mui.com/material-ui/guides/pickers-migration/ and applying our color styles

### DIFF
--- a/src/forms/Form/Form.stories.tsx
+++ b/src/forms/Form/Form.stories.tsx
@@ -143,7 +143,7 @@ export const Playground = (): ReactElement => {
 	const prepopulateValues = object("Prepolulate values", {
 		"textField": "Text field prepopulated",
 		"textArea": "Text area prepopulated",
-		"date": new Date(),
+		"date": new Date().toString(),
 		"check": [
 			"label_1",
 			"label_2"
--- a/src/forms/Form/Form.stories.tsx
+++ b/src/forms/Form/Form.stories.tsx
@@ -143,6 +143,7 @@ export const Playground = (): ReactElement => {
 	const prepopulateValues = object("Prepolulate values", {
 		"textField": "Text field prepopulated",
 		"textArea": "Text area prepopulated",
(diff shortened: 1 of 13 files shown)


src/forms/FormFieldDate/DatePicker/DatePicker.tsx
```diff
@@ -28,44 +22,35 @@ const DatePicker = (props: MosaicFieldProps<any>): ReactElement => {
...
-					/>
-				</DatePickerWrapper>
-			</ThemeProvider>
-		</MuiPickersUtilsProvider>
+		<LocalizationProvider dateAdapter={AdapterDateFns}>
+			<DatePickerWrapper error={!!error} isPickerOpen={isPickerOpen}>
+				<DatePicker
+					renderInput={renderInput}
+					inputFormat="MM/dd/yyyy"
+					value={value}
```
mauriciodeleonc, line 43: Same comment as with the time picker

mauriciodeleonc (changes requested):

src/forms/Form/Form.stories.tsx
```diff
@@ -143,6 +143,7 @@ export const Playground = (): ReactElement => {
 	const prepopulateValues = object("Prepolulate values", {
 		"textField": "Text field prepopulated",
 		"textArea": "Text area prepopulated",
+		"date": new Date(),
```
mauriciodeleonc, line 146: I think the Date issue might still be present, here is what the date looks like on the form state 
image

and this is the hour on the field
image

It says it should be 17:56 pm but since we're on a GMT-5 time zone the actual component says 12:56, this is what shouldn't happen. Formstate and component should be the same.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #98, opened 2022-05-19
MOS-652: Fix additional actions in the Saved Views drawer

JAndresHJGlobant:
## What's included?

- The additional actions displayed on the Saved Views drawer are now shown. The problem was the `z-index` of the Drawer, the actions were being displayed correctly but behind the drawer

## Screenshots
image

### DIFF
--- a/src/components/Drawer.jsx
+++ b/src/components/Drawer.jsx
@@ -12,6 +12,10 @@ const DrawerContent = styled.div`
 	font-size: 14px;
 `
 
+const MUIDrawerStyled = styled(MUIDrawer)`
+	z-index: 1000;
+`
+
 function Drawer(props) {
 	const [state, setState] = useState({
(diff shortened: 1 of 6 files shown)


src/components/Drawer.jsx
```diff
@@ -12,6 +12,10 @@ const DrawerContent = styled.div`
 	font-size: 14px;
 `
 
+const MUIDrawerStyled = styled(MUIDrawer)`
+	z-index: 1000;
```
mauriciodeleonc, line 16: Let's make this a bit larger than the z-index of the top component because then the drawers gets under it and prevents users from saving, closing, etc. 
image

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #112, opened 2022-05-31
MOS-667: Fixes button styles

JAndresHJGlobant:
# What's included?
- Removal of the “large” button format. 
- All non-icon buttons should be 14px in size.
- Font weight set to 500
- Add letter-spacing of 1px to all buttons.
- The medium button should be exactly 36px tall, and the small button should be exactly 30px tall.
- The font-size of icon inside the button was set to 16px
- Removal of the border radius on the hover state for the text buttons
- The black used in many of the buttons was incorrect, it was changed to #1A1A1A

### DIFF
--- a/src/components/Button/Button.styled.ts
+++ b/src/components/Button/Button.styled.ts
@@ -46,25 +46,25 @@ export const ButtonWrapper = styled.span`
   &.normalButton.size_small > button .icon_left {
     margin-right: 4px;
     margin-left: -4px;
-		font-size: 14px;
+		font-size: 16px;
   }
 
   &.normalButton.size_small > button .icon_right {
     margin-left: 4px;
     margin-right: -4px;
-		font-size: 14px;
+		font-size: 16px;
   }
 
   &.normalButton.size_medium > button .icon_left {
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.styled.ts
```diff
@@ -47,39 +43,28 @@ export const ButtonWrapper = styled.span`
...
 
-  &.normalButton.size_large.variant_outlined > button {
-    padding-top: 7px;
-    padding-bottom: 7px;
-  }
-
   &.normalButton.size_small > button .icon_left {
     margin-right: 4px;
     margin-left: -4px;
+		font-size: 14px;
```
mauriciodeleonc, line 49: Everything looks good, the only thing that should be changed is this, it should be 16px:
"When we render an icon inside the button (not the icon button variant but when the button has an icon and text string), the icon should have a 16px font-size. This will shrink the icon in the button a little bit from it’s current point. It should be 16px in both small and medium."

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #120, opened 2022-06-07
MOS-672: Summary page content

JAndresHJGlobant:
## What's included?
- Content component
- Storybook stories
- Unit tests
- Teal icon button variant

### DIFF
--- a/src/components/Content/Content.stories.tsx
+++ b/src/components/Content/Content.stories.tsx
@@ -1,0 +1,73 @@
+import * as React from "react";
+import { ReactElement } from "react";
+import { boolean, withKnobs, text } from "@storybook/addon-knobs";
+import { Meta } from "@storybook/addon-docs/blocks";
+import PictureAsPdfIcon from "@mui/icons-material/PictureAsPdf";
+
+// Components
+import Content, { ContentProps } from ".";
+
+export default {
+	title: "Components/Content",
+	decorators: [withKnobs],
+} as Meta;
+
+const edit = () => alert("Edit button clicked");
+
+const content: ContentProps["content"] = [
+	[
+		{ type: "paragraph", label: "something", value: "Cum quod earum." },
+		{ type: "labelValue", label: "Label", value: "value" },
+	],
+	[
+		{ type: "labelValue", label: "Label", value: "value" },
+		{ type: "labelValue", label: "Label", value: "value" },
+	],
+	[
+		{ type: "labelValue", label: "Label", value: "value" },
+		{ type: "labelValue", label: "Label", value: "value" },
+	],
+	[
+		{ type: "labelValue", label: "Label", value: "value" },
+		{ type: "labelValue", label: "Label", value: "value" },
+	],
+	[{ type: "tags", label: "Tags Label", value: ["Tag", "Tag"] }],
+	[
+		{
+			type: "paragraph",
+			label: "something",
+			value:
+        "Cum quod earum saepe veniam. Quis quis est assumenda in at tempora iusto dicta tempore. Vel mollitia voluptatem pariatur. Cum quod earum saepe veniam. Quis quis est assumenda in at tempora iusto dicta tempore. Vel mollitia voluptatem pariatur.",
+		},
+		{
+			type: "file",
+			label: "NameFile.pdf",
+			value: "· 00/00/0000 at 00:00",
+			icon: PictureAsPdfIcon,
+		},
+	],
+	[
+		{
+			type: "paragraph",
+			label: "something",
+			value:
+        "Cum quod earum saepe veniam. Quis quis est assumenda in at tempora iusto dicta tempore. Vel mollitia voluptatem pariatur. Cum quod earum saepe veniam. Quis quis est assumenda in at tempora iusto dicta tempore. Vel mollitia voluptatem pariatur.",
+		},
+	],
+];
+
+const onClickAdd = () => alert("Add button clicked");
+
+export const Playground = (): ReactElement => {
+	const title = text("Title", "Main Section Title");
+	const showEdit = boolean("Show edit button", true);
+
+	return (
+		<Content
+			title={title}
+			content={content}
+			onClickEdit={showEdit && edit}
+			onClickAdd={onClickAdd}
+		/>
+	);
+};
--- a/src/components/Content/Content.styled.tsx
+++ b/src/components/Content/Content.styled.tsx
@@ -1,0 +1,103 @@
+import styled from "styled-components";
+import theme from "../../theme";
+import EditIcon from "@mui/icons-material/Edit";
+
+export const MainWrapper = styled.div`
+	font-family: ${theme.fontFamily};
+	border-bottom: 2px solid ${theme.colors.gray200};
+`;
+
+export const TitleWrapper = styled.div`
+	align-items: center;
+	display: flex;
+	justify-content: space-between;
+`;
+
+export const Title = styled.span`
+	color: ${theme.colors.almostBlack};
+	font-size: 16px;
+	font-weight: ${theme.fontWeight.semiBold};
+`;
+
+export const StyledEditIcon = styled(EditIcon)`
+	color: ${theme.colors.gray600};
+	cursor: pointer;
+
+	&.MuiSvgIcon-root {
+		border-right: 1px solid ${theme.colors.gray200};
+		padding-right: 16px;
+		width: 16px;
+	}
+`;
+
+export const ButtonsWrapper = styled.div`
+	align-items: center;
+	display: flex;	
+
+	.MuiButton-root {
+		padding-left: 16px;
+	}
+
+	.iconButton {
+		border-right: 1px solid ${theme.colors.gray200};
+		padding-right: 16px;
+	}
+`;
+
+// Styles for the paragraph content type
+export const Paragraph = styled.p`
+	color: ${theme.colors.gray700};
+	font-size: 14px;
+	margin: 0 0 16px;
+`;
+
+// Styles for the LabelValue content type
+
+export const LabelValueWrapper = styled.div`
+	margin-bottom: 16px;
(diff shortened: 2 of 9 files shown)


src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+	const contentForDisplay = showOrHide ? content : content.slice(0, 3);
+
+	/**
+	 * Renders the content of type "labelValue"
+	 * @param contentItem 
+	 * @returns The content with a sctructure of label: value
+	 */
+	const labelValueContent = (contentItem: ContentType) => (
+			<LabelValueWrapper>
```
mauriciodeleonc, line 46: Codacy has a fix for the issue: [Expected indentation of 2 tabs but found 3.](https://app.codacy.com/gh/simpleviewinc/sv-mosaic/pullRequest?prid=9571838)
```suggestion
		<LabelValueWrapper>
```

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+				<Label>{contentItem.label}:</Label>
+				<Value>{contentItem.value}</Value>
```
mauriciodeleonc, line 48: Codacy has a fix for the issue: [Expected indentation of 3 tabs but found 4.](https://app.codacy.com/gh/simpleviewinc/sv-mosaic/pullRequest?prid=9571838)
```suggestion
			<Value>{contentItem.value}</Value>
```

src/components/Content/Content.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const Playground = (): ReactElement => {
+	const title = text("Title", "Main Section Title");
+	const showEdit = boolean("Show edit button", true);
+
+	return (
+		<Content
+			title={title}
+			content={content}
+			onClickEdit={showEdit && edit}
+			onClickAdd={onClickAdd}
```
mauriciodeleonc, line 70: I think we should add a knob or something to make the onClickAdd actually do something, since it always has content, this will never be shown. Maybe a boolean knob that says "clear content" or "show as empty"

src/components/Content/Content.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+		expect(getAllByTestId("content-row").length).toBe(4);
+		expect(getByText("Fourth paragraph")).toBeDefined();
+	});
+});
+
+describe("Content componenent when no content is passed", () => {
+	beforeEach(() => {
+		render(
+			<Content title="Main Section" content={[]} onClickAdd={onClickAdd} />
```
mauriciodeleonc, line 76: I think we should add a test for when content = null or undefined because it's not required... I believe it should produce the same output as here, but just to be sure.

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+	ValueFile,
+	ContentRow,
+	ContentColumn,
+	ParagraphWrapper,
+} from "./Content.styled";
+import EditIcon from "@mui/icons-material/Edit";
+import AddIcon from "@mui/icons-material/Add";
+
+const Content = (props: ContentProps): ReactElement => {
+	const { title, onClickEdit, content, onClickAdd } = props;
```
mauriciodeleonc, line 28: Recommendation (you don't have to change this), I feel it looks / reads better "onEdit" and "onAdd" because "on" already talks about the action.

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+	const [showOrHide, setShowOrHide] = useState(content.length < 3);
```
mauriciodeleonc, line 29: Tip (you don't have to change this), you could leave the variable just as "show" since it's a boolean, show: false means that it's hidden (also I feel it's easier to read).

src/components/Content/Content.styled.tsx
```diff
@@ -0,0 +1,107 @@
...
+	margin: 0;
+`;
+
+export const ParagraphWrapper = styled.div`
+	margin-bottom: 16px;
+`;
+
+// Styles for the LabelValue content type
+
+export const LabelValueWrapper = styled.div`
```
mauriciodeleonc, line 60: Isn't this the same as the paragraph wrapper?

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+	};
+
+	const contentForDisplay = showOrHide ? content : content.slice(0, 3);
+
+	/**
+	 * Renders the content of type "labelValue"
+	 * @param contentItem 
+	 * @returns The content with a sctructure of label: value
+	 */
+	const labelValueContent = (contentItem: ContentType) => (
```
mauriciodeleonc, line 45: Isn't this the same as the paragraph content?

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+	/**
+	 * Renders the content of type "paragraph"
+	 * @param contentItem 
+	 * @returns A paragraph with the value of the content
+	 */
+	const paragraphContent = (contentItem: ContentType) => (
+		<ParagraphWrapper>
+			<Label>{contentItem.label}:</Label>
+			<Paragraph>{contentItem.value}</Paragraph>
```
mauriciodeleonc, line 60: I believe this is the only thing that changes between the labelValueContent and the paragraphContent, so you could reduce the repeated code by just creating a single element and passing a "type" prop to the styled component and that's where you decide what margin and font-size you apply.

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+	const fileContent = (contentItem: ContentType) => {
+		const FileIcon = contentItem?.icon;
+
+		return (
+			<>
+				<LabelWrapper>
+					{FileIcon && <FileIcon />}
+					<Label>{contentItem.label}:</Label>
+				</LabelWrapper>
+				<ValueFile>{contentItem.value}</ValueFile>
```
mauriciodeleonc, line 101: Same comment as with the paragraph.

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+		);
+	};
+
+	return (
+		<MainWrapper>
+			<TitleWrapper>
+				<Title>{title}</Title>
+				<ButtonsWrapper>
+					{content?.length === 0 ? (
+						<Button color="teal" variant="icon" mIcon={AddIcon} onClick={onClickAdd}></Button>
```
mauriciodeleonc, line 112: onClickAdd is optional, so this will still render even if the prop is not passed, which shouldn't happen.

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+						onClickEdit && (
+							<Button
+								color="gray"
+								variant="icon"
+								mIcon={EditIcon}
+								onClick={onClickEdit}
+							></Button>
+						)
+					)}
+					<Button
```
mauriciodeleonc, line 123: I believe this button will still render even if content.length <= 3

src/components/Content/Content.styled.tsx
```diff
@@ -0,0 +1,107 @@
...
+	}
+`;
+
+export const ValueFile = styled(Value)`
+	font-size: 12px;
+	margin-left: 24px;
+`;
+
+export const ContentRow = styled.div`
+  display: flex;
```
mauriciodeleonc, line 101: There are some weird tabs / spaces here and on width: 100%

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,167 @@
...
+					></Button>
+				</ButtonsWrapper>
+			</TitleWrapper>
+			<div>
+				{contentForDisplay.map((items, idx) => {
+					return (
+						<ContentRow data-testid={"content-row"} key={`${idx}-contentLayout`}>
+							{items.map((contentItem, idx) => {
+								if (
+									(contentItem.type === "tags" &&
```
mauriciodeleonc, line 137: All of this logic looks good, but I feel it might be better to process it outside of the render in something like a useEffect so we don't even render all the other components if this throws an error.

mauriciodeleonc (changes requested):

src/components/Content/Content.styled.tsx
JAndresHJGlobant: You are right, I will use the same wrapper for both.

src/components/Content/Content.tsx
JAndresHJGlobant: In this case, it is not. The ParagraphContent renders a `<p>` element whereas the LabelValue are `<span>` elements.

src/components/Content/Content.tsx
JAndresHJGlobant: This cannot be done because of what I explained in a previous comment, they are different HTML elements.

src/components/Content/Content.tsx
JAndresHJGlobant: This can be done since both are `<span>` I removed the ValueFile styled component and use the Value with a prop that defines the styles for each case.

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,196 @@
...
+const Content = (props: ContentProps): ReactElement => {
+	const { title, onEdit, content, onAdd } = props;
+	const [show, setShow] = useState(content?.length < 3);
+
+	/**
+	 * Validates if the combinations of the type of content and
+	 * value are allowed.
+	 */
+	useEffect(() => {
+		if(!content) return;
```
mauriciodeleonc, line 41: Codacy has a fix for the issue: [Expected space(s) after "if".](https://app.codacy.com/gh/simpleviewinc/sv-mosaic/pullRequest?prid=9571838)
```suggestion
		if (!content) return;
```

mauriciodeleonc (approved):

PR #122, opened 2022-06-08
MOS-680: Styles for sideNav component

cristianvj23:
# What's include:

- Same gray for hovering and selected. (only thing that changes when selected is yellow bar and bold).
- Show pointer when hovering over elements.
- Changed component Name to “SideNav”.


src/components/SideNav/SideNav.tsx
```diff
@@ -1,18 +1,18 @@
 import * as React from "react";
 import { ReactElement, useState } from "react";
-import { NavigationProps } from ".";
+import { SideNavProps } from ".";
 import {
 	LinkWrapper,
-	SideNav,
+	SideNavStyle,
```
mauriciodeleonc, line 6: Tip (you don't have to change this), the standard for styled components is "Styled_NameOfComponent_" in this case it would be something like StyledSideNav

mauriciodeleonc (approved):

PR #129, opened 2022-06-13
MOS-682: Fix 'any' in the DateFilter

cristianvj23:
# What's include:

- Added bold tags to Bodytext for DataViewPrimaryFilter
- Added required validation to filterSingleSelect

### DIFF
--- a/src/components/FilterSingleSelect/FilterSingleSelect.stories.tsx
+++ b/src/components/FilterSingleSelect/FilterSingleSelect.stories.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { useState, ReactElement } from "react";
 import FilterSingleSelect from "./FilterSingleSelect";
-import { boolean, withKnobs } from "@storybook/addon-knobs";
+import { boolean, select, withKnobs } from "@storybook/addon-knobs";
 
 export default {
 	title : "Components/FilterSingleSelect",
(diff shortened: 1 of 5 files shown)


src/components/FilterSingleSelect/FilterSingleSelect.tsx
```diff
@@ -17,6 +17,8 @@ export default function FilterSingleSelect(props: FilterSingleSelectProps): Reac
 		options: []
 	});
 
+	if (props.required && !props.data.value) throw new Error("A value is required")
```
mauriciodeleonc, line 20: I'd suggest adding a more complete error message like "Invalid use-case, a value is required but none was provided" so developers know where and how to fix their error.

src/components/FilterSingleSelect/FilterSingleSelect.stories.tsx
```diff
@@ -14,6 +16,9 @@ const options = [
 ]
 
 export const example = (): ReactElement => {
+
+	const required = boolean("Required", false);
```
mauriciodeleonc, line 20: Please add a knob for changing the value / default value, because then whoever is looking at the storybook will never be able to see how the required logic works, they will only see the error message (unless they select an option and then click on the required knob). I'll also recommend adding a kitchen sink story with both types of filter, a required and a non required.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #134, opened 2022-06-15
MOS-674: Popover

JAndresHJGlobant:
## What's included?
- Popover component 
- Storybook examples based on https://xd.adobe.com/view/91b10cca-d20f-48d4-9afc-6fe6f5b06ed5-c4e2/screen/00e4c753-5751-4a62-b826-8580fb99e285/
- Unit tests

### DIFF
--- a/src/components/Popover/Popover.styled.tsx
+++ b/src/components/Popover/Popover.styled.tsx
@@ -1,0 +1,70 @@
+import styled from "styled-components";
+import theme from "@root/theme";
+import Paper from "@mui/material/Paper";
+
+export const StyledPaper = styled(Paper)`
+	&.MuiPaper-root {
+		background-color: white;
+		border: 1px solid ${theme.colors.gray200};
+		max-width: 620px;
+		border-radius: 5px;
+		box-shadow: 0px 2px 6px #1A1A1A26;
+	}
+`;
+// Reference:
+// https://github.com/Patil2099/material-ui/blob/89d9e960cdba5a84a3416fdc0c76bafce1053053/docs/src/pages/components/popper/ScrollPlayground.js
+
+export const PopperSx = {
+	zIndex: 2000,
+	"&[data-popper-placement*='bottom']": {
+		"& .MuiPopper-arrow": {
+			top: 0,
+			left: 0,
+			marginTop: "-0.9em",
+			width: "3em",
+			height: "1em",
+			"&::before": {
+				borderWidth: "0 1em 1em 1em",
+				borderColor: `transparent transparent ${theme.colors.white} transparent`,
+			},
+		},
+	},
+	"&[data-popper-placement*='top']": {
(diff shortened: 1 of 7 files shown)


src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,92 @@
...
+						sx={{
+							boxShadow: "none",
+						}}
+					>
+						<ClickAwayListener onClickAway={onClose}>
+							<StyledPaper>
+								<Arrow ref={setArrowRef} className="MuiPopper-arrow" />
+								<ContentWrapper>
+									<div>{topContent}</div>
+									{bottomContent && <StyledHr data-testid="bottom-content-split" />}
```
mauriciodeleonc, line 80: I think this HR should only render if we have both top and bottom contents, because this means I can pass only bottom and it'll still render

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,92 @@
...
+} from "./Popover.styled";
+
+// Components
+import Popper from "@mui/material/Popper";
+import Grow from "@mui/material/Grow";
+import ClickAwayListener from "@mui/material/ClickAwayListener";
+import Paper from "@mui/material/Paper";
+
+const Popover = (props: PopoverProps): ReactElement => {
+	const {
```
mauriciodeleonc, line 19: I recommend throwing an error if no anchorEl was passed saying something like "Popover component requires an anchor element to position itself but none was provided", and also another error if top and bottom content are not passed saying something like "Popover component expected some content but none was provided".

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,92 @@
...
+					<Paper
+						sx={{
+							boxShadow: "none",
+						}}
+					>
+						<ClickAwayListener onClickAway={onClose}>
+							<StyledPaper>
+								<Arrow ref={setArrowRef} className="MuiPopper-arrow" />
+								<ContentWrapper>
+									<div>{topContent}</div>
```
mauriciodeleonc, line 79: I feel these divs (top and bottom) shouldn't render at all if they were not provided.

src/components/Popover/Popover.styled.tsx
```diff
@@ -0,0 +1,70 @@
...
+				borderWidth: "1em 1em 0 1em",
+				borderColor: `${theme.colors.white} transparent transparent transparent`,
+			},
+		},
+	}
+};
+
+export const Arrow = styled.div`
+	position: absolute;
+  font-size: 7px;
```
mauriciodeleonc, line 49: Not sure why on github there aren't tabs or spaces from line 49 to line 51.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #135, opened 2022-06-15
MOS-677: Adding column tests

carol-bohorquez:
**### What's include:**
* Column test cases (add, remove, order)
* Remove some not used "await"


automation_testing/tests/dataview/data_view.spec.ts
```diff
@@ -1,28 +1,32 @@
...
 
 	test.beforeEach(async ({ page }) => {
 		dataviewPage = new DataviewPage(page);
+		columns = dataviewPage.columnsComponent;
+
 		await dataviewPage.visit();
 	});
 
-	test("Create New", async () => {
+	test("DataView - Create New", async () => {
```
mauriciodeleonc, line 17: I feel this might be repeated as the describe of the test already is "Data View", maybe we could remove it and keep it as it was... or was there a specific reason for you to add it?

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #136, opened 2022-06-15
MOS-683: Actions need to accept an array

JAndresHJGlobant:
## What's included?
- Actions show attribute accepts and processes an array.
- Adds the type to the show actions attribute of: `(DataViewActionShow | boolean)[]`
- Unit tests to the filterAction helper function

### DIFF
--- a/src/components/DataView/utils/bulkActionsUtils.ts
+++ b/src/components/DataView/utils/bulkActionsUtils.ts
@@ -12,17 +12,21 @@ type FilterActionArgs = RowArgs | CheckedArgs;
 type Action = DataViewAction | DataViewAdditionalAction | DataViewBulkAction;
 
 export const filterAction = (action: Action, args: FilterActionArgs) => {
+	if (action.show === undefined) {
+		return true;
+	} 
+
 	const shows = isArray(action.show) ? action.show : [action.show];
 
-	for (const show of shows) {
-		if (show === undefined) {
-			return true;
-		} else if (typeof show === "boolean") {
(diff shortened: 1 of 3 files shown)


src/components/DataView/utils/bulkActionsUtils.ts
```diff
@@ -10,13 +12,17 @@ type FilterActionArgs = RowArgs | CheckedArgs;
...
-		return true;
-	} else if (typeof action.show === "boolean") {
-		return action.show;
-	} else if (typeof action.show === "function") {
-		return action.show(args);
-	} else {
-		throw new Error(`Action ${action.name}.show must be boolean or a function`);
+	const shows = isArray(action.show) ? action.show : [action.show];
+
+	for (const show of shows) {
```
mauriciodeleonc, line 17: I don't think this for of loop is what we're looking for to solve the ticket. When using return statements inside a for loop, the loop will stop running when the first return gets executed. In this case, we want to run over all the array, if every element returns true, then it means we can show the button (action) if at least one of them returns false then we shouldn't render the button. That's why the "every" function can help us, look at this example: 
image

On the first approach it prints true because it returned on the first element, while the second approach returns false because not all the elements return true.

You can test this on the DataViewKitchenSink.tsx file, line 519, you can add the show property there, then in storybook, if you put a false on the show, but it still renders then it's not working as expected.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #140, opened 2022-06-21
MOS-271: Disables duplicated saved views

JAndresHJGlobant:
## What's included?
- Disables saving views with the same name


mauriciodeleonc:
After talking with Owen it was decided that this was a product-specific solution and that we should come with a generic error-handling solution.

PR #142, opened 2022-06-22
MOS-700: Value of filters should not be capitalized

JAndresHJGlobant:
## What's included?
- Value of filters is no longer capitalized

### DIFF
--- a/src/components/DataViewPrimaryFilter.jsx
+++ b/src/components/DataViewPrimaryFilter.jsx
@@ -42,6 +42,7 @@ const LabelWrapper = styled.div`
 	& > .filter-value {
 		font-weight: ${theme.fontWeight.normal};
 		margin-right: 8px;
+		text-transform: none;
 	}
 
 	&.type_optional > * {
--- a/src/components/DataViewPrimaryFilter.jsx
+++ b/src/components/DataViewPrimaryFilter.jsx
@@ -33,10 +33,15 @@ const LabelWrapper = styled.div`
 		background-color: ${theme.colors.blue}45;
 	}
 
-	& > .filter {
+	& > .filter-label {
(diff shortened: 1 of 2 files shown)


src/components/DataViewPrimaryFilter.jsx
```diff
@@ -33,12 +33,17 @@ const LabelWrapper = styled.div`
...
+	& > .filter-label {
 		font-weight: ${theme.fontWeight.normal};
 		margin-right: 8px;
 		text-transform: capitalize;
 	}
 
+	& > .filter-value {
+		font-weight: ${theme.fontWeight.normal};
+		margin-right: 8px;
+	}
```
mauriciodeleonc, line 45: This should have a "text-transform: none" because then it doesn't actually do what the ticket asks. Currently we have a text transform: uppercase on all buttons, but on the filters, we want our values to render just as the user writes them.

src/components/DataViewPrimaryFilter.jsx
```diff
@@ -75,8 +80,8 @@ function DataViewPrimaryFilter(props) {
 				type_${props.type}
 			`}
 		>
-			<BodyText className="filter">{props.label}:</BodyText>
-			<BodyText className="filter"><b>{props.value || "Any"}</b></BodyText>
+			<BodyText className="filter-label">{props.label}:</BodyText>
+			<BodyText className="filter-value"><b>{props.value || "Any"}</b></BodyText>
 			{
 				props.type === "optional" &&
 				<CloseIcon
```
mauriciodeleonc, line 87: Please also on this branch change the variable valueString from "any" to "Any" on file "FilterDate.tsx"

mauriciodeleonc (changes requested):

PR #143, opened 2022-06-24
MOS-686 - Basic Filter tests

carol-bohorquez:
**### What's included:**
* Keyword title and categories filters
* Pages class to move common methods


automation_testing/pages/DataViewPage.ts
```diff
@@ -143,4 +140,45 @@ export class DataviewPage {
...
+	async validateContainsMoreThanOneKeyword(titles: string[], keywords: string[]): Promise<void> {
+
+		let isContaining = false;
+		for (const title of titles) {
+			for (const keyword of keywords) {
+				if (title.toLowerCase().includes(keyword.toLowerCase())) {
+					isContaining = true;
+				}
+			}
+			expect(isContaining, `Expected contains: '${keywords.toString()}' but was '${title}'`).toBe(true);
```
mauriciodeleonc, line 159: I feel this is likely to fail on some scenarios since the "isContaining" flag is not being changed back to false on each iteration, for example let's say that titles[0] does include keywords[0] then isContaining changes to true, it executes the expect, but the flag is still true, then by next iteration, even if titles[0] does not contain keywords[1] it will still mark it as true.

I see 2 alternatives, either add an "else" that changes isContaining to false, or changing it back to false at the start of each iteration.

mauriciodeleonc (changes requested):

automation_testing/pages/DataViewPage.ts
```diff
@@ -143,4 +140,46 @@ export class DataviewPage {
...
+		for (const title of titles) {
+			expect(title.toLowerCase()).toContain(keyword.toLowerCase());
+		}
+	}
+
+	async validateContainsMoreThanOneKeyword(titles: string[], keywords: string[]): Promise<void> {
+
+
+		for (const title of titles) {
+			let isContaining = false;
```
mauriciodeleonc, line 154: I don't think having the flag here fixes the issue. Let's assume titles has a length of 3, and keywords has a lenght of 2: titles[0] includes keywords[0] then the flag is changed to true, but then even if titles[0] doesn't include keywords[1] it will still be true because the flag resets to false until it finishes with all the keywords on title[0]. Let me know if we need to have a meeting to explain it a bit better.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #144, opened 2022-06-24
Mos 670: Top component

cristianvj23:
# What's include:

- Developed component for Summary page top component
- SummaryPageTopComponent types
- SummaryPageTopComponent storybook
- SummaryPageTopComponent styles


src/components/Button/Button.tsx
```diff
@@ -25,7 +25,9 @@ function Button(props: ButtonProps) {
 
 	const addAttrs: MosaicObject = {};
 
-	const MyButton = types[`${props.color}_${props.variant}`];
+	const textLink = props.variant === "text" && props.href ? "_link" : "";
```
mauriciodeleonc, line 28: All of this is actually not needed, you already have everything you need on props and attrs. Also, when a prop.href get's passed, the Button component is smart enough to change the element from <button> to <a> and that means the only changes you have to do are on the css.

src/components/Button/Button.tsx
```diff
@@ -25,7 +25,9 @@ function Button(props: ButtonProps) {
 
 	const addAttrs: MosaicObject = {};
 
-	const MyButton = types[`${props.color}_${props.variant}`];
+	const textLink = props.variant === "text" && props.href ? "_link" : "";
+
+	const MyButton = types[`${props.color}_${props.variant}${textLink}`];
```
mauriciodeleonc, line 30: Same comment as above

src/components/Button/Button.styled.ts
```diff
@@ -471,6 +471,24 @@ export const types = {
 	    line-height: ${textButtonStyles.lineHeight};
     }
   `,
+	black_text_link: styled(BlackOnWhite)`
```
mauriciodeleonc, line 474: As I mentioned in the comment from Button.tsx file, this whole new style is not necessary since the component already makes the transformations enough to what you need. I made some small changes and this is what I ended up needing to style it the same way: 
image

image
This is for styling the icon, is at the beginning of the file.]

Also, it seems that the original style you have is missing a space between the icon and the text but with these new changes I believe you won't need to add them.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #149, opened 2022-06-27
MOS-709: Top component styling issues

JAndresHJGlobant:
## What's included?
- Replaces Title margin-bottom with a margin-top applied to the optional description
- Replaces the bottom margin applied to the container of the Title and Description with a margin-top to the FormNav
- Conditionally rendering the Description
- Modifies style script since with the flag of `--max-warnings=0` the unit tests were failing.

### DIFF
--- a/src/forms/FormNav/FormNav.styled.tsx
+++ b/src/forms/FormNav/FormNav.styled.tsx
@@ -125,6 +125,7 @@ export const FormNavRow = styled.div`
 
 export const FormNavWrapper = styled.div`
   background: white;
+  margin-top: 30px;
   overflow: hidden;
   position: sticky;
   position: -webkit-sticky;
--- a/src/forms/TopComponent/Utils/TitleWrapper.tsx
+++ b/src/forms/TopComponent/Utils/TitleWrapper.tsx
@@ -15,7 +15,7 @@ export const FormTitle = styled.span`
 
   color: ${theme.colors.almostBlack};
   font-size: ${pr => pr.type && pr.type === "drawer" ? "20px" : "28px"};
-  margin: ${pr => pr.type && pr.type === "drawer" ? "auto" : "0 0 8px 0"};
+  margin: ${pr => pr.type && pr.type === "drawer" ? "auto" : ""};
 
   @media (max-width: ${theme.breakpoints.mobile}) {
     margin: 0;
@@ -25,6 +25,7 @@ export const FormTitle = styled.span`
 const Description = styled.span`
   color: ${theme.colors.gray600};
   font-size: 14px;
+  margin-top: 8px;
 
   @media (max-width: ${theme.breakpoints.mobile}) {
     align-self: center;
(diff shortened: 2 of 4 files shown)


src/forms/FormNav/FormNav.styled.tsx
```diff
@@ -125,6 +125,7 @@ export const FormNavRow = styled.div`
 
 export const FormNavWrapper = styled.div`
   background: white;
+  margin-top: 30px;
```
mauriciodeleonc, line 128: On big desktop this is leaving a gap between the top component and the sidenav, but on the xd they look together. 
image

If this affects other places let's just keep the margin-top 30px on all sizes except big desktop (if it doesn't affect then I'd recommend removing it).

src/forms/TopComponent/Views/DesktopView.tsx
```diff
@@ -20,13 +20,12 @@ const BIG_SCREEN_BREAKPOINT = BREAKPOINTS.topComponent.bigScreenView + "px";
 
 const DesktopViewColumn = styled(StyledColumn)`
   justify-content: space-between;
-  padding: 24px 20px 0px 20px;
+  padding: 24px 20px 30px 20px;
```
mauriciodeleonc, line 23: This padding 30px makes some weird spaces, specially on the "Form with Layout" example, because it renders the tabs with a lot of space under them. I'd recommend 2 approaches: 1. Return this to 0 and keep the margin-bottom: 30 on the DesktopTitleActionsRow (I've been reading the ticket and it seems the only issue was when no description was added), or 2. Add a conditional to add those 30px when no sections are being propped down.

mauriciodeleonc (changes requested):

src/forms/TopComponent/Utils/TitleWrapper.tsx
```diff
@@ -11,25 +11,25 @@ const BIG_SCREEN_BREAKPOINT = BREAKPOINTS.topComponent.bigScreenView + "px";
...
 	font-family: ${pr => !pr.type && theme.museoFont};
-	font-weight: ${pr => pr.type && pr.type === "drawer" && "medium"};
+	font-weight: ${pr => pr.type && pr.type === "DRAWER" && "medium"};
 
-  color: ${theme.colors.almostBlack};
-  font-size: ${pr => pr.type && pr.type === "drawer" ? "20px" : "28px"};
-  margin: ${pr => pr.type && pr.type === "drawer" ? "auto" : ""};
+	color: ${theme.colors.almostBlack};
+	font-size: ${pr => pr.type && pr.type === "DRAWER" ? "20px" : "28px"};
+	margin: ${pr => pr.type && pr.type === "DRAWER" && "auto"};
```
mauriciodeleonc, line 18: I believe here, if type !== "DRAWER" the margin will be something like margin: undefined; which won't "really" affect the css but shouldn't be there. 

I suggest if possible, changing to something like: ${pr => pr.type === "DRAWER" && magin: auto; font-weight: "medium";}
 (you could add the font-weight here as they deal with the same logic)

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #151, opened 2022-06-28
MOS-692: Type of number field

JAndresHJGlobant:
## What's included?
- Saves the input value like a number in the form state when its type is number
- Adjusts unit tests affected
- Adds storybook examples


src/forms/FormFieldText/FormFieldText.stories.tsx
```diff
@@ -78,6 +80,7 @@ export const Playground = (): ReactElement => {
 			multiline,
 			helperText,
 			instructionText,
+			type
```
mauriciodeleonc, line 83: Please let's also add a type number example to the KitchenSink

src/forms/Form/Form.stories.tsx
```diff
@@ -142,6 +142,7 @@ export const Playground = (): ReactElement => {
 	const showSections = boolean("Show sections", false);
 	const prepopulateValues = object("Prepolulate values", {
 		"textField": "Text field prepopulated",
+		"numberField": 4596,
```
mauriciodeleonc, line 145: I don't think we need to add a number example here, since the playground only has the most basic example of each field. Here we would have 2 variants of the same field.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #155, opened 2022-06-29
MOS-714 - [TAE] Test Form Layout Story

torres-nahuel:
### Jira Ticket
https://simpleviewtools.atlassian.net/browse/MOS-714

### **What's Included**
* Added tests for From with Layout 
* Renamed Base Page*


automation_testing/pages/Form/FormWithLayoutPage.ts
```diff
@@ -0,0 +1,91 @@
...
+	async getSectionsTitles():Promise<string[]> {
+		const numberOfSections = await this.getNumberOfSectionsFromTopComponent();
+		const sectionTitles = [];
+		for (let i = 0; i <  numberOfSections; i++) {
+			sectionTitles.push(await this.sectionContainer.locator("h1").nth(i).textContent());
+		}
+		return sectionTitles;
+	}
+
+	async getSectionsDescriptions():Promise<string[]> {
```
mauriciodeleonc, line 53: I feel the code in this function is repeated a lot and we have an area of opportunity if we make it generic, I feel instead of having the functions getSectionsTitlesFromTopComponent, getSectionsTitles, getSectionsDescriptions, and getNumberOfColumsInEachRowFromSections we could have something like getSectionsData that receives as argument the type of data to find

For example:
```
	async getSectionsData(container: locator, element: string):Promise<string[]> {
                   ...all your logic
                   for loop
                           container.locator(element)...
        }
```

That way we can make this work with whatever might change / be added in the future.

automation_testing/tests/form/form_with_layout.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+		expect(numberOfSections).toBe(expectedNumberOfTitles);
+		expect(await formWithLayoutPage.getSectionsTitles()).not.toBe("");
+	});
+
+	test("Validate that each Section has a Description", async () => {
+		expect(await formWithLayoutPage.getSectionsDescriptions()).not.toBe("");
+	});
+
+	test("Validate that each row in each section does not have more than 3 columns", async () => {
+		const numberOfColumnsPerRow = await formWithLayoutPage.getNumberOfColumsInEachRowFromSections()
```
mauriciodeleonc, line 24: This line seems to be missing a ";" that might cause issues on our github actions.

automation_testing/tests/form/form_with_layout.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+		for (let i = 0; i < numberOfColumnsPerRow.length; i++) {
+			expect(numberOfColumnsPerRow[i]).toBeLessThanOrEqual(3);
+		}
+	});
+
+	test("Validate that when scrolling to the bottom of the page, the correct Section is selected in Top Component", async () => {
```
mauriciodeleonc, line 30: All tests seem pretty good to me, the only thing I suggest would be to change our titles to be a bit shorter, for example this test could be something like: Correct section selected when scrolling to the bottom of the page.

mauriciodeleonc (changes requested):

automation_testing/pages/Form/FormWithLayoutPage.ts
torres-nahuel: Created getSectionData method to address this.

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #158, opened 2022-06-30
Mos-701: Dataview grid view

cristianvj23:
#What's include?

- Added media queries with the theme brakpoints

### DIFF
--- a/src/components/DataView/DataViewDisplayGrid.jsx
+++ b/src/components/DataView/DataViewDisplayGrid.jsx
@@ -42,11 +42,20 @@ const StyledDiv = styled.div`
 			grid-template-columns: repeat(2, 1fr);
 		};
 
+		@media (min-width: 768px) {
+			grid-template-columns: repeat(3, 1fr);
+		};
+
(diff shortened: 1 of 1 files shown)


src/components/DataView/DataViewDisplayGrid.jsx
```diff
@@ -30,9 +34,20 @@ const StyledDiv = styled.div`
 
 	& > .grid {
 		display: grid;
-		grid-template-columns: repeat(5, 1fr);
+		grid-template-columns: repeat(1, 1fr);
```
mauriciodeleonc, line 37: Okay so I've been playing a lot with the sizes and I think we should add 2 more because the jump between 4 columns to 2 feels odd. After researching common breakpoints I think it's best if we add 1 every 300px approx, so it will look something like this:
image
Here our biggest amount of columns will be 6 and will descend back to 1. I think leaving the biggest as 5 might not be the best because that screen size it's difficult to get on laptops.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #161, opened 2022-07-05
MOS-708: Allow developer to control form buttons

JAndresHJGlobant:
## What's included?
- Remove submission of forms via registration
- Update every story that was registering submit functions
- Update documentation
- Updates unit tests
- Fixed buttons in TopComponents removed and replaced with a list of buttons provided where the Form component is used
- Adds "show" mechanics to conditionally render the TopComponent buttons

### DIFF
--- a/src/forms/Form/Form.stories.tsx
+++ b/src/forms/Form/Form.stories.tsx
@@ -28,6 +28,7 @@ import { DropdownSingleSelectionDef } from "../FormFieldDropdownSingleSelection"
 import { FormFieldChipSingleSelectDef } from "../FormFieldChipSingleSelect";
 import { FormFieldCheckboxDef } from "../FormFieldCheckbox";
 import { TextAreaDef } from "../FormFieldTextArea";
+import { ButtonProps } from "@root/components/Button";
 
 export default {
 	title: "Components/Form",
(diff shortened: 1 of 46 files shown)


src/forms/Form/Form.tsx
```diff
@@ -82,12 +74,10 @@ const Form = (props: FormProps) => {
 							title={title}
 							type={type}
 							description={description}
-							onSubmit={(e) => submit(e)}
-							submitButtonAttrs={submitButtonAttrs}
 							onCancel={onCancel ? (e) => cancel(e) : null}
-							cancelButtonAttrs={cancelButtonAttrs}
 							sections={sections}
 							view={view}
+							buttons={buttons}
```
mauriciodeleonc, line 80: I feel we could filter the buttons here, that way we don't care if it's a form drawer or a regular form, or if the user resizes, the filteredButtons stay the same and it's only processed once. It could be inside a useMemo that only updates when buttons change.

src/forms/Form/Form.stories.tsx
```diff
@@ -125,9 +126,33 @@ const onCancel = () => {
 	alert("Cancelling form, going back to previous site");
 };
 
+const onSubmit = async (dispatch) => {
```
mauriciodeleonc, line 129: This onSubmit and renderButtons seem to be repeated a lot. I think it's better if we create a "storyUtils" file on the utils folder, there we could just export the renderButtons function, that way if we need to change it for all stories we only change it once.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #166, opened 2022-07-11
MOS-702: Simplify form contract

JAndresHJGlobant:
## What's included?

- Removes fields registration from the Form contract
- Docs updated
- Fields are now registered within the Form component

### DIFF
--- a/src/forms/Form/Form.tsx
+++ b/src/forms/Form/Form.tsx
@@ -31,6 +31,17 @@ const Form = (props: FormProps) => {
 	const { view } = useWindowResizer(type);
 
 	useEffect(() => {
+		const registerFields = async () => {
+			await dispatch(
+				formActions.init({
+					fields
+				})
+			);
+		}
+		registerFields();
+	},[fields])
+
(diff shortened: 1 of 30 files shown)


src/forms/Form/Form.tsx
```diff
@@ -30,6 +30,17 @@ const Form = (props: FormProps) => {
 
 	const { view } = useWindowResizer(type);
 
+	useEffect(() => {
+		const registerFields = async () => {
```
mauriciodeleonc, line 34: The only issue I might see with this piece of code is the something similar to what's mentioned in https://simpleviewtools.atlassian.net/jira/software/c/projects/MOS/boards/82?modal=detail&selectedIssue=MOS-716 . Although they're not the same issue I feel it might also cause some memory leaks when unmounting and the async function still trying to be solved. I'd recommend adding something similar to the solutions in that ticket.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #169, opened 2022-07-12
MOS-721: Update MUI to latest and change Pickers import path

JAndresHJGlobant:
What's included?
- Update mui package to the latest version
- Changes import path of the Pickers components and tools


package.json
```diff
@@ -17,9 +17,10 @@
 		"@material-ui/core": "^4.12.3",
 		"@material-ui/icons": "^4.11.3",
 		"@material-ui/pickers": "3.2.10",
-		"@mui/icons-material": "^5.6.2",
+		"@mui/x-date-pickers": "^5.0.0-beta.0",
+		"@mui/icons-material": "^5.8.4",
 		"@mui/lab": "^5.0.0-alpha.80",
```
mauriciodeleonc, line 22: If we're not using lab any more anywhere on the project please let's remove it.

mauriciodeleonc (changes requested):

package.json
JAndresHJGlobant: You're right it's no longer in use

PR #178, opened 2022-07-19
MOS-724 - [TAE] FormFieldText

torres-nahuel:
### Jira Ticket:
https://simpleviewtools.atlassian.net/browse/MOS-724

### What's Included
- Added tests for the FormFieldText and all of its variants inside a Form.


automation_testing/pages/FormFields/FormFieldTextPage.ts
```diff
@@ -0,0 +1,65 @@
...
+
+	async visitPage(): Promise<void> {
+		await this.visit(this.page_path, this.title);
+	}
+
+	async getAutogeneratedText(length: number): Promise<string> {
+		return generateRandomId(length);
+	}
+
+	async getLimitOfMaxChar(): Promise<number> {
```
JAndresHJGlobant, line 50: Typescript is complaining that the value of `this.maxCharCounter.textContent()` could be possibly null. If you are 100% sure that this value will never be null you can use the non-null assertion operator to remove this warning. https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-
If there is a chance that this value is equal to null, you add a check for this case.

```
(property) FormFieldTextPage.maxCharCounter: Locator
Object is possibly 'null'.ts(2531)
```

image

### Solution
image

automation_testing/tests/Components/dataview/save_as.spec.ts
```diff
@@ -1,9 +1,9 @@
 import { test, expect } from "@playwright/test";
-import { dataview_data, saveAs_data } from "../../utils/data/dataview_data";
-import { DataviewPage } from "../../pages/DataView/DataViewPage";
-import { SaveAsComponent } from "../../pages/DataView/SaveAsComponent";
-import { PaginationComponent } from "../../pages/DataView/PaginationComponent";
-import { ColumnsComponent } from "../../pages/DataView/ColumnsComponent";
+import { dataview_data, saveAs_data } from "../../../utils/data/dataview_data";
```
JAndresHJGlobant, line 2: I found that these files that are marked in red need to update some import paths. 

image

image

automation_testing/utils/helpers/helper.ts
```diff
@@ -58,4 +58,14 @@ const getDateFormatted = (date: Date) : string => {
...
+	let newId = "";
+	const characters = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+	const charactersLength = characters.length;
+	for (let i = 0; i < length; i++) {
+		newId += characters.charAt(Math.floor(Math.random() * charactersLength));
+	}
+	return newId;
+}
+
+export { sortDatesAsc, sortDatesDesc, randomIntFromInterval, getMonthFromString, addQuotes, addComma, isACorrentDateRange, randomNumber, getDateFormatted, generateRandomId }
```
JAndresHJGlobant, line 71: This is just a recommendation, it's up to you. You add the `export` to each helper function to not have a single export with so manny function. I mean something like this:

```
export const generateRandomId = (length: number): string => {
	let newId = "";
	const characters = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
	const charactersLength = characters.length;
	for (let i = 0; i < length; i++) {
		newId += characters.charAt(Math.floor(Math.random() * charactersLength));
	}
	return newId;
}
```

JAndresHJGlobant (changes requested):

automation_testing/utils/helpers/helper.ts
torres-nahuel: I like this idea, I was thinking of ways to rework the file, so I will apply this.

automation_testing/pages/FormFields/FormFieldTextPage.ts
JAndresHJGlobant: Yes! That will do the work.

JAndresHJGlobant (approved):

PR #190, opened 2022-07-26
MOS-59: Exists and Not Exists - Uncheck selection options for Categories with Comparisons

cristianvj23:
# What's include?
- Added validations to hide checks and chips when EXISTS and NOT EXISTS have been selected from Categories with comprisons component


src/components/internal/DataViewFilterMultiselectDropdownContent.jsx
```diff
@@ -375,7 +375,7 @@ function DataViewFilterMultiselectDropdownContent(props) {
 					{
 						showList &&
 						<CheckboxList
-							checked={state.selected}
+							checked={optionsDisabled ? [] : state.selected}
```
mauriciodeleonc, line 378: Although this solution solves the issue visually, internally the dataview still receives the options selected, which shouldn't happen. Here you can see that even though the selected filter is "exist" it still gives me the ids of the 2 first options: 
image
What I suggest doing here would be updating state.selected to be [] and keeping the previously selected options on another state variable, that way you can retrieve them if you select another comparison back again. I believe this can be solved on this same file but inside the "menuItems" variable.

mauriciodeleonc (changes requested):
Please look at comments

mauriciodeleonc (approved):

PR #192, opened 2022-07-28
Mos-77:  Backslash in string searches and MOS-78: Keyword Filter Error

cristianvj23:
# What's include:
- Changed RegExp validation of keyword filter by string comparison


src/components/DataView/example/DataViewKitchenSink.tsx
```diff
@@ -66,7 +66,7 @@ const processStringFilter = function ({ name, data, output }) {
 	if (data.comparison === "equals") {
 		output[name] = data.value;
 	} else if (data.comparison === "contains") {
-		output[name] = new RegExp(`.*${data.value}.*`, "i");
+		output[name] = data.value;
 	} else if (data.comparison === "not_contains") {
 		output[name] = new RegExp(`^((?!${data.value}).)*$`, "i")
```
mauriciodeleonc, line 71: Not contains also needs to be just a string because then it messes with title with comparison.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #194, opened 2022-08-01
MOS-701: Sections refactor

JAndresHJGlobant:
## What's included?
- Replace the use of classes to select a section with refs
- Display the right scroll button conditionally
- Use the section index instead of the title to get its id
- Replace `window.replace` with `scrollIntoView` using a ref
- CSS logic refactored
- Rename **fieldsLayoutPos** to **rows**
- Update the typeDef for the “fields” from  (string | FieldDef)[ ][ ][ ] to string[ ][ ][ ].
- Debouncing scroll listener function
- Throwing an error message to indicate that an undefined field was declared. “No field declared for field name 'X' at row 1, column 10 in section 0.”

### DIFF
--- a/src/forms/FormNav/FormNav.tsx
+++ b/src/forms/FormNav/FormNav.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
-import { memo, useState, useRef, useCallback, useEffect, ReactElement } from "react";
+import { useState, useRef, useEffect, ReactElement } from "react";
+import { debounce } from "lodash";
 import {
 	FormNavWrapper,
 	FormNavRow,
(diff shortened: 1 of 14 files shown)


src/forms/FormNav/FormNavTypes.tsx
```diff
@@ -5,4 +5,5 @@ export interface Section {
 
 export interface FormNavProps {
   sections: Section[];
+	sectionsRefs?: HTMLDivElement[] | []
```
mauriciodeleonc, line 8: The prop has some weird indentation,  I'll change this here if there's nothing else to change / improve.

src/forms/FormNav/FormNav.tsx
```diff
@@ -43,43 +64,46 @@ const FormNav = (props: FormNavProps): ReactElement => {
...
+			if (sectionsRefs.length === 0) {
+				return;
+			}
 
-		return () => {
-			window.removeEventListener("scroll", navHighlighter);
+			sectionsRefs?.forEach(current => {
+				const currentSection = current as HTMLDivElement;
+				const sectionHeight = currentSection?.offsetHeight;
+				const sectionTop = currentSection?.offsetTop - 350;
```
mauriciodeleonc, line 88: Not 100% sure if this is what might be causing the issue but the scrolling doesn't work on big screen, I tried using bigger sections but it doesn't seem to do anything... let's talk about it tomorrow.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #198, opened 2022-08-03
MOS-704: Form submit issue

cristianvj23:
# What's include:
- onBlur was commented and added timer to onchange for validations

### DIFF
--- a/src/forms/Form/Col.tsx
+++ b/src/forms/Form/Col.tsx
@@ -66,7 +66,7 @@ const Col = (props: ColPropsTypes) => {
 		imageUpload: FormFieldImageUpload,
 	}), []);
 
-	const doneTypingInterval = 500;
+	const doneTypingInterval = 300;
 	let typingTimer;
 
 	const sendValidateField = async (curr) => {
@@ -81,7 +81,6 @@ const Col = (props: ColPropsTypes) => {
 	};
 
 	const onChangeMap = useMemo(() => {
-		clearTimeout(typingTimer);
 		return fieldsDef.reduce((prev, curr) => {
 			prev[curr.name] = async function (value) {
 				await dispatch(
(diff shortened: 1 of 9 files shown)


src/forms/Form/Col.tsx
```diff
@@ -66,7 +66,22 @@ const Col = (props: ColPropsTypes) => {
...
+		if (curr.pairedFields)
+			curr.pairedFields.forEach(async pairedField => {
+				await dispatch(
+					formActions.validateField({ name: pairedField })
+				);
+			});
+	};
+
 	const onChangeMap = useMemo(() => {
+		clearTimeout(typingTimer);
```
mauriciodeleonc, line 84: The clearTimeout should go before the setTimeout (line 93) in order to actually clean it, because right now it waits until the "doneTypingInterval" value finishes but it gets executed the same amount of letters the input has. Also I think we can reduce the value from 500 to 300.

src/forms/Form/Col.tsx
```diff
@@ -150,10 +166,10 @@ const Col = (props: ColPropsTypes) => {
 						value={value}
 						error={error}
 						onChange={onChange}
-						onBlur={onBlur}
+						// onBlur={onBlur}
```
mauriciodeleonc, line 169: If we're commenting something out we should make sure it doesn't break anything of what we currently have. If you check the playground story you'll see some components break because we're trying to execute the onBlur. We don't need to pass it, but then we need to comment it also in those other components or add a conditional to not execute them if they're undefined. 
Here are the components that throw the error (there might be more):
image

image

The Image Video Document... doesn't remove the error message when selecting a media type.
image

image

The Table field doesn't remove the error message when selecting a media type.
image

image

mauriciodeleonc (changes requested):

src/forms/Form/Col.tsx
cristianvj23: The errors not related to onBlur, for example with Image Video Document or Table Field, were found to come from earlier branches. They will be resolved anyway.

mauriciodeleonc (approved):

PR #206, opened 2022-08-08
MOS-744: Color picker including opacity in hex representation

JAndresHJGlobant:
## What's included?

- Transformation of the rgba color value to hex

## How to test it?
I used the following tool to check if the rgba color was transformed correctly
https://rgbacolorpicker.com/rgba-to-hex


src/forms/FormFieldColorPicker/FormFieldColorPicker.tsx
```diff
@@ -28,7 +47,7 @@ const FormFieldColorPicker = (
 	
 	const onColorChange = (color: ColorResult) => {
 		setColor(color);
-		onChange(color.hex);
+		onChange(RGBAToHexA(color.rgb));
```
mauriciodeleonc, line 50: This is good for changing the color from the field to the form, but we should also have a mechanic (probably in a useEffect) that processes the value when it goes from the form into the field. In this example I opened the Form Playground, added the value to the color picker: "#a8001791", clicked prepopulate, and this is what it looked like: 
image

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #208, opened 2022-08-09
MOS-743: Added form story with leftNav

cristianvj23:
# What's include
- Added DMS form example story with leftNav component


src/forms/Form/Form.stories.tsx
```diff
@@ -1122,3 +1128,185 @@ export const DefaultValues = (): ReactElement => {
 		</>
 	);
 };
+
+const FakeTopBar = function(props: { variant: string, openNav : MouseEventHandler }): ReactElement {
```
mauriciodeleonc, line 1132: All of this piece of code could be easily reduced if we just export them from LeftNav.stories.tsx, there's no need for this to be copied and pasted here if it's the same as that other file. The only thing that changes between them is the "<Lorem />" but that can easily be just a prop that gets passed down in the "DMS Example story". Let's try to follow the DRY principle as much as possible (Don't Repeat Yourself).

src/forms/Form/Form.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { ReactElement, useEffect, useMemo, useState, useCallback } from "react";
+import { ReactElement, useEffect, useMemo, useState, useCallback, useRef, MouseEventHandler } from "react";
```
mauriciodeleonc, line 2: All of these imports will go away once the change from the comment below takes place.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #209, opened 2022-08-09
Mos 717: Improper required fields for data view

JAndresHJGlobant:
## What's included?
- Check which DataView props are required and mark them as so.

### DIFF
--- a/src/components/DataView/example/DataViewKitchenSink.tsx
+++ b/src/components/DataView/example/DataViewKitchenSink.tsx
@@ -672,7 +672,12 @@ function DataViewKitchenSink(): ReactElement {
 				filter
 			});
 		},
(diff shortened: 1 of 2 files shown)


src/components/DataView/example/DataViewKitchenSink.tsx
```diff
@@ -672,29 +672,29 @@ function DataViewKitchenSink(): ReactElement {
 				filter
 			});
 		},
-		savedViewAllowSharedViewSave
+		savedViewAllowSharedViewSave,
```
mauriciodeleonc, line 675: I'm not sure why you decided to move these props here, but they don't affect so I'll leave them here

mauriciodeleonc (approved):

PR #220, opened 2022-08-23
MOS-765: onBlur errors when using Date and Time Fields

cristianvj23:
# What's include:
- Commented onBlur() from timePicker


src/forms/FormFieldDate/TimePicker/TimePicker.tsx
```diff
@@ -20,7 +20,7 @@ const TimeFieldPicker = (props: MosaicFieldProps<TimePickerDef>): ReactElement =
 
 	const handleOpenState = async () => {
 		setIsPickerOpen(!isPickerOpen);
-		await onBlur();
+		// await onBlur();
```
mauriciodeleonc: The onBlur might come back in the future, so rather than having to comment and uncomment every time we need it, I'd recommend adding a conditional that prevents the onBlur from executing if none is present, as we have in other components: if(onBlur) await onBlur(). Since this is the only thing that needs to be changed I'll update it here.

mauriciodeleonc (approved):

PR #237, opened 2022-09-07
Mos-761: Text default comparison not functional

cristianvj23:
# What's include:
- Refactor to send comparisonDefault prop to DataViewFilterText component
- Unit test case to check it throw an error when developer send an invalid comparison

### DIFF
--- a/src/components/DataView/DataView.test.tsx
+++ b/src/components/DataView/DataView.test.tsx
@@ -12,7 +12,7 @@ describe("DataViewFilterText component", () => {
 				label = "Title with Comparisons"
 				type = "optional"
 				data = {{}}
-				comparisonDefault = "" 
+				comparisonDefault = ""
 				args = {{
 					comparisons: ["equals", "not_equals", "contains", "not_contains", "exists", "not_exists"]
 				}}
@@ -32,7 +32,7 @@ describe("DataViewFilterText component", () => {
 				label = "Title with Comparisons"
 				type = "optional"
 				data = {{}}
-				comparisonDefault = "not_equals" 
+				comparisonDefault = "not_equals"
 				args = {{
 					comparisons: ["equals", "not_equals", "contains", "not_contains", "exists", "not_exists"]
 				}}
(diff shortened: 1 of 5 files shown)


src/components/DataView/DataView.test.tsx
```diff
@@ -64,4 +64,41 @@ describe("DataViewFilterText component", () => {
 		const filter = await screen.findByText("Equals");
 		expect(filter).toBeTruthy()
 	})
-});
\ No newline at end of file
+
+	it("Should throw an error when the developer has passed an invalid comparisonDefault prop", async () => {
+		/* await act(async () => {
```
mauriciodeleonc, line 69: If we're not using this anymore let's remove it

src/components/DataViewFilterText/DataViewFilterText.tsx
```diff
@@ -119,8 +119,13 @@ function DataViewFilterText(props: DataViewFilterTextProps) {
 	// 	throwOnInvalid : true
 	// });
 
+	if (props.comparisonDefault && validComparisons.find( validComparison => validComparison.value === props.comparisonDefault) === undefined)
+		throw new Error("The selected comparison is not a valid comparison")
+
 	const [anchorEl, setAnchorEl] = useState(null);
-	const comparison = props.data.comparison || (props.comparisonDefault || "equals");
+	const comparison = props.args && props.args.comparisons
```
mauriciodeleonc, line 126: I think this is confusing and can be cleaned a lot. I'm not sure what the difference is between props.data.comparison || (props.comparisonDefault || "equals") : props.data.comparison || "equals"; it seems that can be rethinked.

src/components/DataViewFilterText/DataViewFilterText.tsx
```diff
@@ -119,8 +119,13 @@ function DataViewFilterText(props: DataViewFilterTextProps) {
...
 	// });
 
+	if (props.comparisonDefault && validComparisons.find( validComparison => validComparison.value === props.comparisonDefault) === undefined)
+		throw new Error("The selected comparison is not a valid comparison")
+
 	const [anchorEl, setAnchorEl] = useState(null);
-	const comparison = props.data.comparison || (props.comparisonDefault || "equals");
+	const comparison = props.args && props.args.comparisons
+		? props.data.comparison || (props.comparisonDefault || "equals")
+		: props.data.comparison || "equals";
```
mauriciodeleonc, line 128: I don't know where to put this but I was poking around trying to see if it worked for all components using DataViewFilterText... it does but only if the filter is optional (see DataViewKitchenSink), for primary filters it doesn't seem to work. I'm guessing this has something to do with some useEffects or memoization or something like that but let's talk about it tomorrow.

mauriciodeleonc (changes requested):

src/components/DataViewFilterText/DataViewFilterText.tsx
```diff
@@ -119,8 +118,23 @@ function DataViewFilterText(props: DataViewFilterTextProps) {
...
 	// });
 
+	if (props.args.comparisonDefault && validComparisons.find( validComparison => validComparison.value === props.args.comparisonDefault) === undefined)
+		throw new Error("The selected comparison is not a valid comparison");
+
+	const getComparison = () => {
+		if (props.args && props.args.comparisons) {
+			if (props.data.comparison) {
+				return props.data.comparison;
+			} else if (props.args.comparisonDefault) {
```
mauriciodeleonc, line 128: There's one small thing missing here... the ifs say: If I have comparisons, then if I have a comparison selected by the user, then return it, if I don't and I have a comparisonDefault then return it... but this is not contemplating that the comparisons that developers pass might not contain comparisonDefault. A practical example:
props.args.comparisons = ['contains', 'not_contains'],
props.args.comparisonDefault = 'equals'

This means that our filterText component will "select" the equals comparison but it isn't even present in the list of available comparisons, so it will break. I suggest doing something like this in line 128 else if (props.args.comparisonDefault && props.args.comparisons.includes(props.args.comparisonDefault))
After that, if that's not the case then default to "equals"

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #238, opened 2022-09-12
MOS-777: Update Content mechanics

JAndresHJGlobant:
## What's included?
Updates the Content component with the following features:

- Use an array of fieldDefs to know what to render using a transform function
- Adds getValues function that resolves the values that will fill the fields
- Sections to define where the fields should be rendered. These are based on columns with a maximum of 2
- Throws an error when an undefined field is passed to sections.
- Throws an error if a section with more than 2 columns is defined

Additional tasks:
- Refactors ColorPicker component to abstract the container that shows the selected color
- Creates transform_chips( ) and transform_colorPicker( )
- Unit tests created for both the Content component updates and the new transforms


src/components/Content/Content.styled.ts
```diff
@@ -34,72 +55,44 @@ export const StyledEditIcon = styled(EditIcon)`
...
-	font-size: 14px;
+export const FieldContainer = styled.div`
+	margin-bottom: 24px;
 `;
 
-// Styles for the tags content type
+// Transforms styles
 export const ChipsWrapper = styled.div`
-	margin: 24px 0 40px;
+	div {
```
mauriciodeleonc, line 83: I think the chips shouldn't have the 8px, only the label (which I think it already has)

src/components/Content/Content.tsx
```diff
@@ -1,143 +1,150 @@
...
+			}
+
+			if (currentField?.show && !showContent(currentField?.show)) {
+				return;
+			}
+
+			return currentField?.transforms.map((transform) => (
+				<FieldContainer key={currentField.name}>
+					<Label>{currentField.label}</Label>
+					{transform({ data: values[currentField?.column ? currentField?.column : currentField.name] })}
```
mauriciodeleonc, line 136: We should wrap this in a div or span that passes the styles that are required in the figma (font-size: 14px, and color; gray600). I already tested this and it should only affect texts.

src/components/Content/Content.styled.ts
```diff
@@ -34,72 +55,44 @@ export const StyledEditIcon = styled(EditIcon)`
 
 export const ButtonsWrapper = styled.div`
 	align-items: center;
-	display: flex;	
+	display: flex;
 
 	.MuiButton-root {
 		padding-left: 16px;
```
mauriciodeleonc, line 61: The hover state looks weird when using padding because it makes the blue background to grow on the left side. I suggest changing it to margin-left: 8px (because the button already has 8 to have a total of 16)

mauriciodeleonc (changes requested):

src/components/Content/Content.stories.tsx
```diff
@@ -1,81 +1,129 @@
...
+	const singleColumn = boolean("Single column", false);
+	const showChips = boolean("Show chips", true);
+
+	const fieldDef: ContentFieldDef[] = [
+		{
+			name: "chips",
+			label: "Chips using transform_chips()",
+			transforms: [transform_chips()],
+			column: "tags",
+			show: [showChips, () => showChips]
```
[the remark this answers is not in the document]
JAndresHJGlobant: This is just to demonstrate that the show function could contain a boolean or a function that returns a boolean,

mauriciodeleonc (approved):

PR #243, opened 2022-09-21
MOS-779 - Optimize automation tests

torres-nahuel:
### Jira Ticket:
[<link>](https://simpleviewtools.atlassian.net/browse/MOS-779)

### What's Included:
- The objective of this PR is to optimize the automation suite to run tests faster in the same browser context and parallel wherever possible.


automation_testing/tests/FormFields/formFieldImageVideoLinkDocumentBrowsing.spec.ts
```diff
@@ -105,7 +115,7 @@ test.describe("FormFields - FormFieldTable - Kitchen Sink", () => {
 		await ffImageVideoLinkDocumentBrowsingPage.documentButton.click();
```
mauriciodeleonc, line 115: This is the only error I'm getting when running the tests: 
image

automation_testing/tests/Components/dataview/data_view.spec.ts
```diff
@@ -1,48 +1,67 @@
...
+	test.beforeEach(async() => {
+		await page.reload();
+	});
+
+	test.afterAll(async ({ browser }) => {
+		browser.close;
+	});
+
+	test("Create New", async ({ page }) => {
+		page.on("dialog", async dialog => {
```
mauriciodeleonc, line 27: I think this dialog listener can be reduced to a function that just gets called with the message it might expect. We were previously doing something similar with "setDialogValidationListener" is there a reason we stopped using that?

mauriciodeleonc (changes requested):

automation_testing/tests/Components/dataview/data_view.spec.ts
torres-nahuel: This method was causing some issues in the test. For example, if I ran all the tests, the "Edit Icon" test will fail because the expected value is "Create New". I don't know why is taking the expected value of the previous test, that's why I chose to give each test its own dialog method.

mauriciodeleonc (approved):

PR #245, opened 2022-09-23
MOS-767: Migrate FormField stories to v6

JAndresHJGlobant:
## What's included?
All the FormFields stories were updated to be compatible with storybook v6

### Notes
- All Form stories except the Playground stories stayed the same because no controls were needed to add.
- The DMS example story was not displaying the content (children) since in the NavWrapper.tsx component the .main CSS class has an ``overflow: hidden;`` It was removed, but it's strange why in our storybook v5 the children were shown despite containing this style.
- Comment this line``#COPY webpack.config.js /app/webpack.config.js`` in the Dockerfile to be able to spin up the container.

### DIFF
--- a/src/forms/FormFieldCheckbox/FormFieldCheckbox.stories.tsx
+++ b/src/forms/FormFieldCheckbox/FormFieldCheckbox.stories.tsx
@@ -1,7 +1,8 @@
 import * as React from "react";
 import { useMemo } from "react";
-import { FormFieldCheckboxDef } from ".";
+import FormFieldCheckbox, { FormFieldCheckboxDef } from ".";
 import { FieldDef } from "@root/components/Field";
+import { ComponentMeta } from "@storybook/react";
 
 // Components
 import Form, { useForm } from "@root/components/Form";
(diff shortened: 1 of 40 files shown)


src/forms/FormFieldCheckbox/FormFieldCheckbox.stories.tsx
```diff
@@ -1,118 +1,116 @@
...
+import { checkboxOptions } from "./FormFieldCheckboxUtils";
+import { excludedFormFieldsControls, onCancel, renderButtons } from "@root/utils/storyUtils";
 
-// export default {
-// 	title: "FormFields/FormFieldCheckbox",
-// 	decorators: [withKnobs],
-// } as Meta;
+export default {
+	title: "FormFields/FormFieldCheckbox",
+	component: Form,
```
mauriciodeleonc, line 15: Almost in all stories I saw that component was Form, but I think we should be using their respective field (in this case FormFieldCheckbox).... the objective would be to fully show the inputSettings in the knobs, for example, for this component we should have "options", just like how we have the formFieldText

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #249, opened 2022-10-04
MOS-794: DataView Filter Single Selection [Design]

cristianvj23:
# What's include?
- Changes to the layout of the DataViewFilterSingleSelection component


src/components/MenuBase/MenuBase.tsx
```diff
@@ -3,15 +3,38 @@ import { createContext, ReactElement } from "react";
...
 
 import { MenuBaseProps, MenuBaseContextProps } from "./MenuBaseTypes";
+import styled from "@emotion/styled";
+import theme from "@root/theme";
 
 export const MenuBaseContext = createContext<MenuBaseContextProps>(null);
 
+const StyledMUIMenu = styled(MUIMenu)`
+	& > div {
+		box-shadow: 0px 2px 6px rgba(26, 26, 26, 0.16);
```
mauriciodeleonc, line 13: This is the same as our "theme.colors.almostBlack" with 16% opacity. Let's change it so if it ever changed in the future we don't have to go around manually updating it.

mauriciodeleonc: If you click on the copy button it will automatically give you the property with the color and the opacity, all you'd need to do is update the color to our almostBlack.
image

src/components/MenuBase/MenuBase.tsx
```diff
@@ -3,15 +3,38 @@ import { createContext, ReactElement } from "react";
...
+		margin-top: 4px;
+
+		& > ul {
```
mauriciodeleonc, line 16: If we observe this style we can see that it's actually not doing anything on the ul, this is a bad practice as it will make the component calculate a style that's just empty: 
image

For future components what you could do is simply using regular combinators / selectors: https://www.w3schools.com/css/css_combinators.asp

E.g. & ul .MuiTouchRipple-child {} This way the style will only be added to the MuiTouchRipple element.

mauriciodeleonc: Here we're assuming that MenuBase will always receive as children a ul, but in reality MenuBase can receive any ReactNode / JSX Element, so this css won't work for all cases, also it changes the style for whatever child element developers might pass. It's always a good practice to have each component deal internally with their own things. In this case, we know we want those styles to apply to the of MenuBase, so they should be responsibility of the children... here the children we want to modify is MenuItem, so it should have those styles.

src/components/MenuBase/MenuBase.tsx
```diff
@@ -3,15 +3,38 @@ import { createContext, ReactElement } from "react";
...
+			& .MuiTouchRipple-child {
+				background-color: ${theme.colors.grayHover};
+			}
+			& > li {
+				padding: 8px 16px;
+				&:active {
```
mauriciodeleonc, line 22: Here the &:active is also an empty class (same as the ul comment from above)

src/components/MenuItem/MenuItem.tsx
```diff
@@ -8,12 +8,27 @@ import theme from "@root/theme";
...
-	min-width: 150px;
 
 	&.MuiMenuItem-root {
 		min-height: 42px;
 	}
 
+	&.css-kk1bwy-MuiButtonBase-root-MuiMenuItem-root.Mui-selected {
+		background-color: ${theme.colors.white};
+		.menuLabel {
+			font-weight: 700;
```
mauriciodeleonc, line 19: We already have a font-weight for this in our theme file. Let's try to prevent repeating our code as it will save us time in the future.

src/components/MenuItem/MenuItem.tsx
```diff
@@ -8,12 +8,27 @@ import theme from "@root/theme";
...
 
 const StyledMenuItem = styled(MUIMenuItem)`
-	min-width: 150px;
 
 	&.MuiMenuItem-root {
 		min-height: 42px;
 	}
 
+	&.css-kk1bwy-MuiButtonBase-root-MuiMenuItem-root.Mui-selected {
+		background-color: ${theme.colors.white};
```
mauriciodeleonc, line 17: I don't think we really need to specify the background-color for just when it's selected. We could take this to the style from line 12 and that way it will always stay white.

mauriciodeleonc (changes requested):

src/components/MenuItem/MenuItem.tsx
```diff
@@ -8,12 +8,27 @@ import theme from "@root/theme";
...
+		.menuLabel {
+			font-weight: 700;
+		}
+		&:hover {
```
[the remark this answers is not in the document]
cristianvj23: Yes, the styles are being overwritten. I will add !important

src/components/MenuItem/MenuItem.tsx
cristianvj23: Ok, I have to use !important because by default MUI adds a background on the selected element

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #250, opened 2022-10-04
MOS-793: Data view filter date design

JAndresHJGlobant:
## What's included?

- Changes the styles of the buttons bar of every DataView filter.

### DIFF
--- a/automation_testing/tests/Components/dataview/advanced_filters.spec.ts
+++ b/automation_testing/tests/Components/dataview/advanced_filters.spec.ts
@@ -430,4 +430,76 @@ test.describe.parallel("Components - Data View - Advanced Filters", () => {
 		await advancedFilters.waitForElementLoad();
 		expect(await pagination.paginationValue.textContent()).toBe(`1-${dataview_data.resultPerPageDefault} of ${dataview_data.totalRecords}`);
 	});
+
+	test("Validate the Created filter styles.", async () => {
+		const expectedFontWeight = "700";
+		const expectedFontSize = "14px";
+		const expectedApplyButtonColor = "(0, 141, 168)";
+		const expectedCleanrAndCancelButtonColor = "(26, 26, 26)";
+
+		await advancedFilters.moreBtn.click();
+		await advancedFilters.createdOption.check();
+		await advancedFilters.applyBtn.click();
+		await advancedFilters.optionalFilters.nth(0).locator("button").click();
+		await advancedFilters.waitForElementLoad();
+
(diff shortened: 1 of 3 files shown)


src/components/DataViewFilterDropdownButtons/DataViewFilterDropdownButtons.styled.tsx
```diff
@@ -0,0 +1,33 @@
...
+  & {
+    display: flex;
+    justify-content: space-between;
+    margin-top: 1rem;
+  }
+`;
+
+export const StyledButton = styled(Button)`
+  &.normalButton > button {
+    font-size: 16px;
```
mauriciodeleonc, line 16: I know the ticket says 16px, but let's keep it as 14... this is something that we've already talked with the design team but they haven't changed their specs.

src/components/DataViewFilterDropdownButtons/DataViewFilterDropdownButtons.styled.tsx
```diff
@@ -0,0 +1,33 @@
...
+`;
+
+export const StyledButtonRow = styled(ButtonRow)`
+  .button + .normalButton {
+    margin: 0;
+  }
+
+  .button:first-child {
+    border-right: 2px solid ${theme.colors.gray200};
+    padding-right: 16px;
```
mauriciodeleonc, line 27: Buttons already have 8px padding, so we should only add 8px extra to sum a total of 16. (same with the padding left of the right child).

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

automation_testing/tests/Components/dataview/advanced_filters.spec.ts
```diff
@@ -430,4 +430,76 @@ test.describe.parallel("Components - Data View - Advanced Filters", () => {
 		await advancedFilters.waitForElementLoad();
 		expect(await pagination.paginationValue.textContent()).toBe(`1-${dataview_data.resultPerPageDefault} of ${dataview_data.totalRecords}`);
 	});
+
+	test("Validate the Created filter styles.", async () => {
```
mauriciodeleonc, line 434: After seeing the difference between these 2 tests, they seem to be exactly the same, the only thing that changes is the "advancedFilters.createdOption" and "advancedFilters.updatedOption". Let's put all in a function that receives the type. For example: async  testOptions(option) {  ... await advancedFilters[option].check(); ...  }

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #253, opened 2022-10-07
MOS-808: Update FormNav tabs logic

cristianvj23:
#What's include?
-Updated form nav logic so tabnav is not displayed when it receives one section
-Added knob to show one or three sections on the form

### DIFF
--- a/src/components/Form/Form.stories.tsx
+++ b/src/components/Form/Form.stories.tsx
@@ -72,6 +72,7 @@ export const Playground = (): ReactElement => {
 	const required = boolean("Required", true);
 	const disabled = boolean("Disabled", false);
 	const showSections = boolean("Show sections", false);
+	const showMultipleSections = boolean("Show multiple sections", true);
 	const prepopulateValues = object("Prepolulate values", {
 		"textField": "Text field prepopulated",
 		"textArea": "Text area prepopulated",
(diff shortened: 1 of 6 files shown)


src/components/Form/Form.stories.tsx
```diff
@@ -72,6 +72,7 @@ export const Playground = (): ReactElement => {
 	const required = boolean("Required", true);
 	const disabled = boolean("Disabled", false);
 	const showSections = boolean("Show sections", false);
+	const showMultipleSections = boolean("Show multiple sections", true);
```
mauriciodeleonc, line 75: Suggestion: We could get rid of this knob and make the "showSections" a select knob with 0, 1, 2, or 3 as options, if 0 then it's like we didn't have options.

src/components/Form/Form.stories.tsx
```diff
@@ -446,10 +450,18 @@ export const Playground = (): ReactElement => {
...
 					description={text("Description", "This is a description example")}
 					state={state}
-					fields={fields}
+					fields={
+						showSections && !showMultipleSections ? oneSectionFields : fields
+					}
 					dispatch={dispatch}
 					getFormValues={loadReady && onLoad}
-					sections={showSections && sections}
+					sections={
```
mauriciodeleonc, line 458: If we change the showSections knob to be a select then we can do something like showSections === 0 ? undefined : sections.slice(0, showSections)

mauriciodeleonc (changes requested):

src/components/Form/Form.stories.tsx
```diff
@@ -282,6 +283,20 @@ export const Playground = (): ReactElement => {
 						],
 					}
 				} as FieldDef<FormFieldRadioDef>,
+				{
```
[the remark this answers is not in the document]
cristianvj23: Yes, if you look at line 428, I'm using .slice() to get the first slice of the array, and it needs the table field to be in the first seven positions of the array.

mauriciodeleonc (approved):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #256, opened 2022-10-17
MOS-812: Update gaps between icon and text buttons

JAndresHJGlobant:
## What's included?
- Increase space between the right or left icon and the button's text.

### DIFF
--- a/automation_testing/pages/BasePage.ts
+++ b/automation_testing/pages/BasePage.ts
@@ -107,8 +107,8 @@ export class BasePage {
 			return true;
 		}
 		return false;
-  }
-  
+	}
+
 	async selectAndDeleteText(stringLenght:number): Promise<void> {
 		await this.page.keyboard.press("ArrowRight");
 		await this.page.keyboard.down("Shift");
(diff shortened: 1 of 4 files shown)


mauriciodeleonc (approved):

automation_testing/pages/BasePage.ts
```diff
@@ -118,4 +118,19 @@ export class BasePage {
...
+
+	async validateFontWeightFromElement(element: Locator, expectedValue: string): Promise<void> {
+		const elementFontWeight = await ((element).evaluate(el => getComputedStyle(el).fontWeight));
+		expect(elementFontWeight).toBe(expectedValue);
+	}
+
+	async validateMarginValueFromElement(element: Locator, expectedValue: string, isRight: boolean): Promise<void> {
+		let elementMargin: string;
+		if (isRight) {
+			elementMargin = await ((element).evaluate(el => getComputedStyle(el).marginLeft));
```
mauriciodeleonc, line 130: I think the conditions might be swapped. If isRight is true shouldn't it be testing for the marginRight, and marginLeft when false?

mauriciodeleonc (changes requested):

automation_testing/pages/BasePage.ts
torres-nahuel: I think the condition is not wrong. If the icon is on the left side of the button, the margin with the text will be on the right side of the icon. That's why I wrote the condition to be that way.

mauriciodeleonc (approved):

PR #261, opened 2022-10-18
MOS-817: Long Column Name in Column Order

cristianvj23:
# What's include?
- Updated styles for column order text label from dataView
- Added an example column with large label text


src/components/DataView/DataViewColumnDrawerContent.tsx
```diff
@@ -41,6 +41,15 @@ const StyledWrapper = styled.div`
 		display: flex;
 		justify-content: space-between;
 		padding-left: 15px;
+
+		& > span {
```
mauriciodeleonc, line 45: This is also affecting the buttons inside the .right > .item as they also have spans inside of them. Please change this to only affect the fist span that is a child of the .right > .item. You can do this with & > span:first-child

mauriciodeleonc (approved):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #263, opened 2022-10-19
MOS-818: Updates to Content Component

JAndresHJGlobant:
# What's included?
- Ability to define an array of buttons to be displayed
- The content is now shown in a row fashion
- Removes getValues function to use only data
- Unit tests updated


src/components/Content/Content.stories.tsx
```diff
@@ -129,32 +154,26 @@ export const Playground = (): ReactElement => {
...
-					resolve(undefined);
-				}
-			}, 500);
-		});
-	};
+	let oneColumToDisplay;
+	let twoColumnsToDisplay;
+	let buttonsToDisplay;
+
+	if (fields && data) {
```
mauriciodeleonc, line 161: Fields and data are not being changed and always exist so this will always be true. I believe we could change this to be moved to a function that gets called in line 174 (in sections = {}) that way we can also reduce the amount of code in that single line to make it easier for others to read / understand.

src/components/Content/Content.tsx
```diff
@@ -174,50 +111,52 @@ const Content = (props: ContentProps): ReactElement => {
...
-					{canShowMore && values && (
-						<Button
-							color="teal"
-							variant="text"
-							label={showMore ? "Less Details" : "More Details"}
-							onClick={showDetails}
-						></Button>
-					)}
+				<ButtonsWrapper>
+					{buttons.map((button, idx) => (
```
mauriciodeleonc, line 140: Here we're assuming buttons will always exist, but since they're optional this might throw an "can't map undefined" error. Also I believe the <ButtonsWrapper> should only render if buttons exist.

src/components/Content/Content.tsx
```diff
@@ -174,50 +111,52 @@ const Content = (props: ContentProps): ReactElement => {
...
-					{rightColumn && rightColumn.length > 0 && (
-						<ContentColumn>
-							{renderColumn(rightColumn, "right")}
-						</ContentColumn>
-					)}
-				</ColumnsContainer>
-			)}
+			<div>
+				{data && sectionsToRender.map((section, idx) => {
+					if (section?.length >= 3) {
```
mauriciodeleonc, line 147: I believe this error could be thrown in the sectionsToRender or even in a useEffect all the way on the top, that way it breaks without having the code have to process everything up until here.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #267, opened 2022-10-20
MOS-804: Not exists filter returns all records including those with title

cristianvj23:
# What's include?
- Fixed exists and not_exists filters in tittle with comparisons filter

### DIFF
--- a/src/components/DataView/example/DataViewKitchenSink.tsx
+++ b/src/components/DataView/example/DataViewKitchenSink.tsx
@@ -61,7 +61,7 @@ const api = new JSONDB(mappedData, {
 });
 
 const processStringFilter = function ({ name, data, output }) {
-	if (data.value === undefined) { return; }
+	if (data.comparison !== "exists" && data.comparison !== "not_exists" && data.value === undefined) { return; }
 
 	if (data.comparison === "equals") {
 		output[name] = data.value;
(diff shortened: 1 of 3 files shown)


src/components/DataView/example/DataViewKitchenSink.tsx
```diff
@@ -61,7 +61,7 @@ const api = new JSONDB(mappedData, {
 });
 
 const processStringFilter = function ({ name, data, output }) {
-	if (data.value === undefined) { return; }
+	if (data.comparison !== "exists" && data.comparison !== "not_exists" && data.value === undefined) { return; }
```
mauriciodeleonc, line 64: Suggestion: in this function you're repeating conditionals (comparison exists and comparison not_exists) twice, first in line 64 and then in lines 74 and 76. You could split your code in 2 ifs, one where value is undefined, and one where we have value, that way you only run the conditions where needed. For example:
```suggestion
	if (data.value === undefined) {
		if (data.comparison === "exists") {
			output[name] = { $exists: true };
		} else if (data.comparison === "not_exists") {
			output[name] = { $exists: false };
		}

		return;
	} else {
		if (data.comparison === "equals") {
			output[name] = data.value;
		} else if (data.comparison === "contains") {
			output[name] = { $contains: data.value };
		} else if (data.comparison === "not_contains") {
			output[name] = { $not_contains: data.value };
		} else if (data.comparison === "not_equals") {
			output[name] = { $ne: data.value };
		}
	}
```

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #268, opened 2022-10-21
MOS-826: Updates to ImageVideoDocumentLinkSetup component

cristianvj23:
# What's include?
- Updated imageVideoLinkDocumentBrowsing component

### DIFF
--- a/src/forms/FormFieldImageVideoLinkDocumentBrowsing/FormFieldImageVideoLinkDocumentBrowsing.tsx
+++ b/src/forms/FormFieldImageVideoLinkDocumentBrowsing/FormFieldImageVideoLinkDocumentBrowsing.tsx
@@ -160,39 +160,45 @@ const FormFieldImageVideoLinkDocumentBrowsing = (
 		<ImageVideoLinkDocumentBrowsingContainer>
 			{(Array.isArray(value) && value?.length === 0) || !value ? (
 				<BrowsingContainer>
-					<BrowseSpan>
+					<BrowseSpan
+						className = { hasOptions ? "has-options" : "" }
+					>
 						{!hasOptions ? "No browsing options" : "Browse:"}
 					</BrowseSpan>
-					<BrowseOptionsContainer>
-						{fieldDef?.inputSettings?.handleSetImage && (
-							<BrowseOption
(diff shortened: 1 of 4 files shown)


src/forms/FormFieldImageVideoLinkDocumentBrowsing/FormFieldImageVideoLinkDocumentBrowsing.tsx
```diff
@@ -160,39 +160,45 @@ const FormFieldImageVideoLinkDocumentBrowsing = (
 		<ImageVideoLinkDocumentBrowsingContainer>
 			{(Array.isArray(value) && value?.length === 0) || !value ? (
 				<BrowsingContainer>
-					<BrowseSpan>
+					<BrowseSpan
+						className = { hasOptions ? "has-options" : "" }
```
mauriciodeleonc, line 164: I don't think BrowseSpan should change it's style depending on whether there are options or not. The one that should be changing is BrowseOptionsContainer since it's optional.

src/forms/FormFieldImageVideoLinkDocumentBrowsing/ImageVideoLinkDocumentBrowsing.styled.tsx
```diff
@@ -53,8 +53,11 @@ export const BrowseSpan = styled.span`
   color: ${theme.colors.gray600};
   font-family: ${theme.fontFamily};
   font-size: 16px;
-  margin-bottom: 24px;
-  margin-top: 43px;
+
+  &.has-options {
```
mauriciodeleonc, line 57: Same comment as in the main file

mauriciodeleonc: In BrowseOptionsContainer (line 17) we no longer need the margin bottom since the parent component has a style of display: flex and justify-content: center. The only thing we should add there is the margin space between the options and the label, which I believe should be 24px.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #274, opened 2022-11-02
MOS-834: Side Nav - Active Updates

cristianvj23:
# What's include?
- Updated default link active on sideNav

### DIFF
--- a/src/components/SideNav/SideNav.stories.tsx
+++ b/src/components/SideNav/SideNav.stories.tsx
@@ -86,14 +86,14 @@ export const Example = (): ReactElement => {
 	];
 
 	let linkLabels: string[] = ["Invalid link"]
-	links.forEach(link => link.forEach(l => linkLabels = [...linkLabels, l.label]))
+	links.forEach(section => section.forEach(link => linkLabels = [...linkLabels, link.label]))
 
 	const parentHeight = number("Parent height (px)", 500);
-	const defaultLink = select("Default Link", linkLabels, linkLabels[2])
+	const active = select("Active Link", linkLabels, linkLabels[2])
 
 	return (
 		<div style={{ display: "flex", height: parentHeight }}>
-			<SideNav links={links} defaultLink={defaultLink} />
+			<SideNav links={links} active={active} />
 			<div>{content}</div>
 		</div>
 	);
--- a/src/components/SideNav/SideNav.stories.tsx
+++ b/src/components/SideNav/SideNav.stories.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { ReactElement, useState } from "react";
-import { withKnobs, number } from "@storybook/addon-knobs";
+import { withKnobs, number, select } from "@storybook/addon-knobs";
 import { Meta } from "@storybook/addon-docs/blocks";
 
 // Components
@@ -20,8 +20,6 @@ export default {
 
 export const Example = (): ReactElement => {
 	const [content, setContent] = useState<JSX.Element>(null);
-	const parentHeight = number("Parent height (px)", 500);
-
 	const links: Link[][] = [
 		[
 			{
@@ -87,9 +85,15 @@ export const Example = (): ReactElement => {
 		],
 	];
 
+	let linkLabels: string[] = ["Invalid link"]
+	links.forEach(link => link.forEach(l => linkLabels = [...linkLabels, l.label]))
+
+	const parentHeight = number("Parent height (px)", 500);
+	const defaultLink = select("Default Link", linkLabels, linkLabels[2])
+
 	return (
 		<div style={{ display: "flex", height: parentHeight }}>
-			<SideNav links={links} />
+			<SideNav links={links} defaultLink={defaultLink} />
 			<div>{content}</div>
 		</div>
 	);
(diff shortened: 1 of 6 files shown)


src/components/SideNav/SideNav.tsx
```diff
@@ -26,6 +26,28 @@ const SideNav = (props: SideNavProps): ReactElement => {
 		link.onClick();
 	};
 
+	const findLinkselected = () => {
+		let linkSelected: Link
```
mauriciodeleonc, line 30:
```suggestion
		let linkSelected: Link;
```

src/components/SideNav/SideNav.tsx
```diff
@@ -26,6 +26,28 @@ const SideNav = (props: SideNavProps): ReactElement => {
 		link.onClick();
 	};
 
+	const findLinkselected = () => {
```
mauriciodeleonc, line 29: Please be careful with the camel casing
```suggestion
	const findLinkSelected = () => {
```

src/components/SideNav/SideNav.tsx
```diff
@@ -26,6 +26,28 @@ const SideNav = (props: SideNavProps): ReactElement => {
...
+			throw new Error("Default link is not in the links array.");
+		} else {
+			return linkSelected
+		}
+	}
+
+	useEffect(() => {
+		if  (links) {
+			if (defaultLink) {
+				setSelectedLink(defaultLink)
```
mauriciodeleonc, line 42: I think this approach might break in some scenarios. For example: Here we're first assigning the default link to the selected link and then we're validating that it exists. I think we should first validate and then (if it exists) assign it.

Also... if we look at what the "onLinkClicked" function does we'll see that there we're already doing a "setSelectedLink" to whatever the parameter is, so here we're doing it twice.

src/components/SideNav/SideNav.stories.tsx
```diff
@@ -87,9 +85,15 @@ export const Example = (): ReactElement => {
...
+	let linkLabels: string[] = ["Invalid link"]
+	links.forEach(link => link.forEach(l => linkLabels = [...linkLabels, l.label]))
+
+	const parentHeight = number("Parent height (px)", 500);
+	const defaultLink = select("Default Link", linkLabels, linkLabels[2])
+
 	return (
 		<div style={{ display: "flex", height: parentHeight }}>
-			<SideNav links={links} />
+			<SideNav links={links} defaultLink={defaultLink} />
```
mauriciodeleonc, line 96: Update to "active"
```suggestion
			<SideNav links={links} active={active} />
```

src/components/SideNav/SideNav.stories.tsx
```diff
@@ -87,9 +85,15 @@ export const Example = (): ReactElement => {
 		],
 	];
 
+	let linkLabels: string[] = ["Invalid link"]
+	links.forEach(link => link.forEach(l => linkLabels = [...linkLabels, l.label]))
+
+	const parentHeight = number("Parent height (px)", 500);
+	const defaultLink = select("Default Link", linkLabels, linkLabels[2])
```
mauriciodeleonc, line 92:
```suggestion
	const active = select("Active Link", linkLabels, linkLabels[2])
```

src/components/SideNav/SideNav.tsx
```diff
@@ -26,6 +26,28 @@ const SideNav = (props: SideNavProps): ReactElement => {
 		link.onClick();
 	};
 
+	const findLinkselected = () => {
+		let linkSelected: Link
+		links.forEach(link => link.forEach(l => l.label === defaultLink ? linkSelected = l : null))
```
mauriciodeleonc, line 31: This is just a suggestion to make the code easier to read.

If we look at the "const links" in SideNav stories, you'll see it looks something like this
`const links = [
    [ //Section 1
        {...myLink1},
        {...myLink2},
    ],
    [ //Section 2
        {...myLink3},
        {...myLink4},
    ]
]
`

This means that the links prop has sections and each section has links, so I'd suggest doing something like this:
```suggestion
		links.forEach(section=> section.forEach(link => link.label === defaultLink ? linkSelected = link : null))
```

src/components/SideNav/SideNav.tsx
```diff
@@ -26,6 +26,28 @@ const SideNav = (props: SideNavProps): ReactElement => {
...
+		if  (links) {
+			if (defaultLink) {
+				setSelectedLink(defaultLink)
+				onLinkClicked(findLinkselected())
+			} else {
+				setSelectedLink(links[0][0].label)
+				onLinkClicked(links[0][0])
+			}
+		}
+	}, [defaultLink])
```
mauriciodeleonc, line 49: This will most probably throw a dependency warning because we should also be including "links" here since it's being used inside the useEffect.
```suggestion
	}, [links, active])
```

src/components/SideNav/SideNav.tsx
```diff
@@ -26,6 +26,28 @@ const SideNav = (props: SideNavProps): ReactElement => {
...
+		links.forEach(link => link.forEach(l => l.label === defaultLink ? linkSelected = l : null))
+		if (!linkSelected) {
+			throw new Error("Default link is not in the links array.");
+		} else {
+			return linkSelected
+		}
+	}
+
+	useEffect(() => {
+		if  (links) {
```
mauriciodeleonc, line 40: If we look at SideNavTypes we'll see that links is required, so this "if" might be redundant as it will always be true.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #275, opened 2022-11-03
MOS-838: Update references to simply gold

JAndresHJGlobant:
## What's included?
Update references to the simplyGold color

Components that were updated
- Checkbox
- Button
- Chip
- RadioButton
- SideNav
- SummaryPageTopComponent
- ToggleSwitch - This won't have opacity

Since the new colors are plain RGB values and not RGBA the background when hovering over the switch thumb will not have opacity. I tried to add the opacity by doing something like the following but the tumb was also affected by the opacity of the background color

```
.MuiSwitch-colorPrimary {
    &.Mui-checked {
      &:hover {
        background-color: ${theme.newColors.simplyGold["100"]};
	opacity: 30%
      }
    }
  }
```

### Previous
image

### Updated
image

### DIFF
--- a/src/components/Button/Button.styled.ts
+++ b/src/components/Button/Button.styled.ts
@@ -209,7 +209,7 @@ const textButtonStyles = {
 export const types = {
 	yellow_contained: styled(BlackOnYellow)`
     & > button {
-      background-color: ${theme.colors.simplyGold};
+      background-color: ${theme.newColors.simplyGold["100"]};
       border-radius: 0;
       font-size: 14px;
       text-transform: uppercase;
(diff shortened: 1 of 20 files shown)


src/components/Button/Button.styled.ts
```diff
@@ -209,20 +209,20 @@ const textButtonStyles = {
...
     .MuiButton-contained.Mui-disabled {
       color: ${theme.colors.almostBlack};
-      background-color: ${theme.colors.simplyGold};
+      background-color: ${theme.newColors.simplyGold["100"]};
       opacity: 0.5;
     }
 
     & > button:hover {
-      background-color: ${theme.colors.simplyGoldHover};
+      background-color: ${theme.newColors.simplyGold["100"]};
```
mauriciodeleonc, line 225: I think this should be another color. Comparing the theme file and the figma, the color for simplyGoldHover is the same as the darkerSimplyGold[100], let's keep it as simplyGoldHover or change it to newColors.darkerSimplyGold['100'] please.

mauriciodeleonc (changes requested):
I noticed the color is also being used manually in other places:

- ToggleSwitch.styled.tsx in lines 51 and 54 (using a hex to rgba converter it looks it should be simplyGold['40']).
- FormFieldChipSingleSelect.test.tsx in line 103 (is using rgb but should be simplyGold['100']).

mauriciodeleonc (approved):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #276, opened 2022-11-03
MOS-840: Replace references of blueTeal with realTeal

JAndresHJGlobant:
## What's included?

Updates references of the blueTeal color, now is using realTeal with its opacity variants.

### Components that were changed:

- Buttons (text variants)
- Label info icon of FormFields
- Text and Links Popover variant
- Snackbar icon
- Date and Time pickers
- FormFieldImageUpload
- ImageVideoLinkDocumentBrowsing

### DIFF
--- a/src/forms/FormFieldImageUpload/FormFieldImageUpload.styled.ts
+++ b/src/forms/FormFieldImageUpload/FormFieldImageUpload.styled.ts
@@ -14,7 +14,7 @@ import CircularProgress from "@mui/material/CircularProgress";
 export const DragAndDropContainer = styled.div`
   align-items: center;
   border: ${(pr) => (pr.isOver ? `1px dashed ${theme.newColors.realTeal["100"]}` : "")};
-  background-color: ${(pr) => (pr.isOver ? theme.newColors.realTeal["10"] : theme.colors.gray200)};
+  background-color: ${(pr) => (pr.isOver ? theme.newColors.realTeal["20"] : theme.colors.gray200)};
   display: flex;
   flex-direction: column;
   height: 204px;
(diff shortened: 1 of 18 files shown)


src/forms/FormFieldImageUpload/FormFieldImageUpload.styled.ts
```diff
@@ -13,8 +13,8 @@ import CircularProgress from "@mui/material/CircularProgress";
 
 export const DragAndDropContainer = styled.div`
   align-items: center;
-  border: ${(pr) => (pr.isOver ? `1px dashed ${theme.colors.teal}` : "")};
-  background-color: ${(pr) => (pr.isOver ? "#008DA81A" : theme.colors.gray200)};
+  border: ${(pr) => (pr.isOver ? `1px dashed ${theme.newColors.realTeal["100"]}` : "")};
+  background-color: ${(pr) => (pr.isOver ? theme.newColors.realTeal["10"] : theme.colors.gray200)};
```
mauriciodeleonc, line 17: Do we have the realTeal with 10% opacity? I think the lowest was 20%. If we don't then please change it to 20

mauriciodeleonc: I just saw that you added the 0.1 to the themes file, but let's keep it in 20 because the figma file only uses multiples of 2

src/theme/theme.ts
```diff
@@ -94,7 +94,7 @@ const newColors: { [key in MY_OBJECT_KEYS]: Color } = {
 	simplyGreen: {}
 };
 
-const opacities = [1, 0.8, 0.6, 0.4, 0.2];
+const opacities = [1, 0.8, 0.6, 0.4, 0.2, 0.1];
```
mauriciodeleonc, line 97: Let's keep it on multiples of 2. If design needs an opacity that's not here they need to update the figma first.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #278, opened 2022-11-07
MOS-831: DataView activeFilters typing

JAndresHJGlobant:
## What's included?

Use an array of strings for the activeFilters prop.

### DIFF
--- a/src/components/DataView/DataViewFilters.tsx
+++ b/src/components/DataView/DataViewFilters.tsx
@@ -9,6 +9,7 @@ import { DataViewFilterMultiselectDropdownContent } from "@root/components/DataV
 import Button from "../Button";
 import theme from "@root/theme";
 import { useMosaicTranslation } from "@root/i18n";
+import { DataViewProps } from "./DataViewTypes";
 
 const StyledDiv = styled.div`
 	margin-bottom: -4px;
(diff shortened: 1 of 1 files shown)


src/components/DataView/DataViewFilters.tsx
```diff
@@ -109,14 +110,14 @@ function DataViewFilters(props: DataViewFiltersProps) {
 		});
 	}
 
-	const onActiveFiltersChange = function(activeFiltersParam) {
+	const onActiveFiltersChange = function(activeFiltersParam: { value: DataViewProps["activeFilters"] }) {
```
mauriciodeleonc, line 113: Everything looks good, the only thing I noticed is that I believe we might want to do something like this:
```suggestion
	const onActiveFiltersChange = function(activeFiltersParam: { value: DataViewProps["activeFilters"], comparison: string }) {
```
Becuase when the `onActiveFiltersChange` function get's propped down to the `onApply`, it gets passed both value and comparison. Right now it might not trigger an error but later when we update the file that uses the `onApply`, it might.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #285, opened 2022-11-15
MOS-872: Update data view title bar

JAndresHJGlobant:
## What's included?
Updates DataViewTitleBar according to Figma: https://www.figma.com/file/Uba9eKQBJklt1YNJLJ3fzd/SimpleView-Components?node-id=5%3A431&t=LoeBPOrxa3uTuGfr-0

## Screenshot
image


mauriciodeleonc (changes requested):
Notes:
- We should move DataViewFilters from DataView.tsx into DataViewTilteBar.tsx, once this is done let's add the paddings specified in figma and the gray border at the bottom.
- Let's add the corresponding types to DataViewTilteBar (there's a high chance you could just reference the ones in DataViewTypes)

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #286, opened 2022-11-16
MOS-869: Update create table actions row

JAndresHJGlobant:
## What's included?

- Creation of the DataViewActionsRow according to Figma https://www.figma.com/file/Uba9eKQBJklt1YNJLJ3fzd/SimpleView-Components?node-id=498%3A6292&t=o9zZ25TLndVMyrAZ-0
- Removes columns control from the Table's headers
- Types DataViewActionsRow props


src/components/DataView/DataViewDisplayList.tsx
```diff
@@ -30,35 +28,19 @@ interface DataViewDisplayListProps {
 	additionalActions?: any;
 	primaryActions?: any;
 	onCheckboxClick?: any;
+	transformedData?: any;
+	activeColumnObjs?: any;
 }
 
 function DataViewDisplayList(props: DataViewDisplayListProps) {
-	// todo validate props
```
mauriciodeleonc, line 36: Looking at DataView.tsx I'm not completely sure why these variables were deleted from here. It seems the only thing that changes is that now that processing it's done in DataView but the values don't seem to be needed anywhere else yet so why not keep them here?

src/components/DataView/DataView.tsx
```diff
@@ -358,21 +367,32 @@ function DataView (props: DataViewProps): ReactElement  {
...
-		return props.limitOptions || [
-			25,
-			50,
-			100
-		]
-	}, [props.limitOptions]);
-
 	const viewContainerRef = useRef(null);
 
+	const activeColumns = useMemo(() => {
```
mauriciodeleonc, line 372: Same comment as in DataViewDisplayList.tsx

src/components/DataView/DataView.tsx
```diff
@@ -389,33 +409,35 @@ function DataView (props: DataViewProps): ReactElement  {
 					onActiveFiltersChange={props.onActiveFiltersChange}
 				/>
 			</div>
-			<div className="headerRow filters">
+			<div className={display === "list" ? "headerList" : "headerRow filters"}>
+				<div className="left">
```
mauriciodeleonc, line 413: I'm not sure why the DataViewFilters were moved here... if you load the storybook you'll see they appear twice 
image

They should only appear in the DataViewTitleBar which is part of the other ticket you were working on.

Also I'm not sure what the .headerList class is needed for.

src/components/DataView/DataView.tsx
```diff
@@ -438,6 +460,8 @@ function DataView (props: DataViewProps): ReactElement  {
 					limit={props.limit}
 					count={props.count}
 					rowCount={props.data.length}
+					transformedData={transformedData}
```
mauriciodeleonc, line 463: Same comment as DataViewDisplayList.tsx

src/components/DataView/DataViewActionsRow/DataViewActionsRow.styled.ts
```diff
@@ -0,0 +1,8 @@
+import styled from "styled-components";
+
+export const DataViewActionsRowWrapper = styled.div`
+	align-items: center;
+	display: flex;
+	justify-content: space-between;
+	margin-bottom: 20px;
```
mauriciodeleonc, line 7: I believe the margin bottom should be part of .headerRow > .right because if we look at the specs, the whole row should have 20px padding all around

src/components/DataView/DataView.tsx
```diff
@@ -27,6 +27,15 @@ const StyledWrapper = styled.div`
 		margin-bottom: 8px;
 	}
 
+	& > .headerList {
+		display: flex;
+		flex-direction: column;
+	}
+
+	& > .headerRow.title {
+		margin-left: 12px;
```
mauriciodeleonc, line 36: Looking at the figma I couldn't find these 12px. The title bar should have 24px all around except at the bottom (it should be 32) but you already added those in the other ticket.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):
Dataview filters failing. This is expected as their locators were moved but a filter ticket is on the way and it'll change styles, behaviors and locators so it would be pointless to update them here now.

PR #288, opened 2022-11-17
MOS-870:  Update DataView table design

cristianvj23:
# What's include?
- Updated DataView table design


mauriciodeleonc (changes requested):
Notes:
- I recommend merging with develop before working on the changes because there are a lot of new styling changes that might help you.
- In the figma specs the table should have 20px padding on the left and right (not on the top and bottom because that's something Andres' already working on).
- The actions column is missing the label "Actions"
- There are some missing paddings between elements, it should be 32px but it's currently 12px

src/components/DataView/DataViewTHead.tsx
```diff
@@ -243,7 +247,7 @@ function DataViewTHead(props: DataViewTHeadProps) {
 					<StyledTh key="_actions" className={`
 						paddingRight
 						${ !props?.bulkActions?.length ? "paddingLeft" : "" }
-					`}>
+					`}> <span className="columnHeader">Actions</span>
```
mauriciodeleonc, line 250: 2 observations here... 1. This should be on the next line to improve readability and 2. The line below already does this. The only thing that needed to change was that it should always render instead of only when the props.onColumnsChange is false. If you look at the DataViewLocales.json file and look for the "actions" property you'll see it already says "Actions" so the text will appear twice.

src/components/DataView/DataViewTHead.tsx
```diff
@@ -61,7 +64,8 @@ const StyledTh = styled.th`
 
 	& > .columnHeader > .icon {
 		visibility: hidden;
-		font-size: 20.58px;
+		height: 21.34px;
```
mauriciodeleonc, line 67: I'm not sure where these sizes come from, the figma has 24 x 24px
image

src/components/DataView/DataViewTHead.tsx
```diff
@@ -15,13 +15,16 @@ import { DataViewProps } from "./DataViewTypes";
...
+		padding: 0px 12px;
+	};
 `
 
 const StyledTh = styled.th`
 	font-size: 14px;
 	text-align: left;
 	font-weight: 400;
-	padding: 5px 0px;
+	padding: 8px 32px 8px 0px;
```
mauriciodeleonc, line 27: We shouldn't add another 32px as the columns already use one with the ".paddingRight" class.
```suggestion
	padding: 8px 0px;

```

src/components/DataView/DataViewTr/DataViewTr.tsx
```diff
@@ -1,27 +1,14 @@
...
+import DataViewTd from "../DataViewTd";
+import DataViewActionsButtonRow from "../DataViewActionsButtonRow";
 import { Draggable } from "react-beautiful-dnd";
 import DragIndicatorIcon from "@mui/icons-material/DragIndicator";
-import { DataViewProps } from "./DataViewTypes";
-import styled from "styled-components";
-import theme from "@root/theme";
+import { TableRow } from "./DataViewTr.styled";
+import { DataViewTrProps } from "./DataViewTrTypes";
 // import { DataViewAction, DataViewAdditionalAction } from "./DataViewTypes";
```
mauriciodeleonc, line 10: Let's delete this commented line if it's not needed anymore.

src/components/DataView/DataViewTr/DataViewTr.tsx
```diff
@@ -1,27 +1,14 @@
...
-	onReorder?: DataViewProps["onReorder"];
-	onCheckboxClick?: any;
-	primaryActions?: any;
-	additionalActions?: any;
-	originalRowData?: any;
-	columns?: any;
-	row?: any;
-	rowIdx: number;
-}
 // interface DataViewTrProps {
```
mauriciodeleonc, line 12: Let's delete this commented line if it's not needed anymore.

src/components/DataView/DataViewTr/DataViewTrTypes.ts
```diff
@@ -0,0 +1,22 @@
...
+import {
+	DataViewAction,
+	DataViewAdditionalAction,
+	DataViewBulkAction,
+	DataViewColumn,
+	DataViewProps
+} from "../DataViewTypes";
+
+export interface DataViewTrProps {
+	bulkActions?: DataViewBulkAction[];
```
mauriciodeleonc, line 12: Most of these are already declared in DataViewProps (just like you have in the "onReorder") 
```suggestion
	bulkActions?: DataViewProps["bulkActions"];
	checked?: boolean;
	onReorder?: DataViewProps["onReorder"];
	onCheckboxClick?: React.MouseEventHandler<HTMLButtonElement>;
	primaryActions?: DataViewProps["primaryActions"];
	additionalActions?: DataViewProps["additionalActions"];
	originalRowData?: MosaicObject;
	columns?: DataViewProps["columns"];
	row?: MosaicObject;
	rowIdx: number;
```

src/components/DataView/DataViewTBody.tsx
```diff
@@ -2,14 +2,15 @@ import React from "react";
...
-import DataViewTr from "./DataViewTr";
+import DataViewTr from "./DataViewTr/DataViewTr";
 import theme from "@root/theme";
 import { MosaicObject } from "@root/types";
 // import { DataViewAction, DataViewAdditionalAction } from "./DataViewTypes";
 
 const StyledTBody = styled.tbody`
 	& > tr {
 		border-bottom: ${theme.borders.lightGray};
+		padding: 8px 12px;
```
mauriciodeleonc, line 13: I don't think the 8px padding should go here, if you use the inspect element over any td it'll show this: 
image

This is giving each element 5px of padding (this is a few lines under this). Also the 12px should only be added to the left of the first td and the right of the last td

src/components/DataView/DataViewTHead.tsx
```diff
@@ -15,13 +15,16 @@ import { DataViewProps } from "./DataViewTypes";
 
 const StyledWrapper = styled.thead`
 	text-align: left;
+	&>tr {
```
mauriciodeleonc, line 18: Same comment as on the TBody, applying the padding to the tr doesn't actually apply the padding as you can see here: 
image

It has to go inside the first and last TD's of the row.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #293, opened 2022-11-23
Mos 856 Address field country/state updates

mauriciodeleonc:
# What's included?
- Updated FormFieldAddress types to now require getOptionsCountries and getOptionsStates callbacks.
- Updated AdressDrawer to get the corresponding country and state options from these callbacks rather than the countries JSON file.
- Updated Form and Address stories and unit tests to include the new props.
- Created a utils file that exports those 2 helper functions to be reused both in stories and unit tests.

### DIFF
--- a/src/forms/FormFieldAddress/Address.test.tsx
+++ b/src/forms/FormFieldAddress/Address.test.tsx
@@ -5,6 +5,7 @@ import {
 	fireEvent,
 	screen,
 	waitFor,
+	act,
 } from "@testing-library/react";
 import "@testing-library/jest-dom";
 import * as React from "react";
@@ -14,6 +15,10 @@ import { ReactElement } from "react";
 import AddressCard from "./AddressCard";
 import { IAddress, AddressFieldDef } from ".";
 import Form, { useForm } from "@root/components/Form";
+
+// Utils
+import { getOptionsCountries, getOptionsStates } from "./utils/optionGetters";
+// import { renderButtons } from "@root/utils/storyUtils";
 
 const address: IAddress = {
 	id: 1,
@@ -52,6 +57,7 @@ export const AddressFormFieldExample = (props: { inputSettings?: AddressFieldDef
 			state={state}
 			fields={newFields}
 			dispatch={dispatch}
+			// buttons={renderButtons(dispatch)}
 		/>
 	);
 };
@@ -63,9 +69,14 @@ const {
 	getAllByRole,
 	queryAllByTestId,
 	queryByText,
+	findAllByTestId,
+	findAllByRole,
+	findByLabelText,
+	findByText,
+	findByTestId,
 } = screen;
 
-const addNewAddress = () => {
+const addNewAddress = async () => {
 	document.createRange = () => ({
 		setStart: jest.fn(),
 		setEnd: jest.fn(),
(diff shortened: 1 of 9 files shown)


src/forms/FormFieldAddress/Address.test.tsx
```diff
@@ -79,47 +90,81 @@ const addNewAddress = () => {
...
+			getOptionsCountries,
+			getOptionsStates,
+		}}/>);
 	});
 
-	it("should add a new address card and then remove it", async () => {
-		expect(queryAllByTestId("address-card-test")).toStrictEqual([]);
-		addNewAddress();
-		fireEvent.click(getByText("Save"));
+	it.only("should add a new address card and then remove it", async () => {
```
JAndresHJGlobant, line 149: Remove the ``.only`` method to run all tests.

src/forms/FormFieldAddress/Address.test.tsx
```diff
@@ -52,6 +57,7 @@ export const AddressFormFieldExample = (props: { inputSettings?: AddressFieldDef
 			state={state}
 			fields={newFields}
 			dispatch={dispatch}
+			// buttons={renderButtons(dispatch)}
```
JAndresHJGlobant, line 60: I try running the tests and have these results:

image

Could it be that the tests are failing since these buttons are not rendered?

src/forms/FormFieldAddress/Address.test.tsx
```diff
@@ -79,47 +90,81 @@ const addNewAddress = () => {
...
-	it("should add a new address card and then remove it", async () => {
-		expect(queryAllByTestId("address-card-test")).toStrictEqual([]);
-		addNewAddress();
-		fireEvent.click(getByText("Save"));
+	it.only("should add a new address card and then remove it", async () => {
+		// expect(await findAllByTestId("address-card-test")).not.toBeInTheDocument();
+		await addNewAddress();
+
+		const form = await findAllByTestId("form-test-id");
+		screen.debug(form[1]);
```
JAndresHJGlobant, line 154: Remove this ``screen.debug`` line to avoid extra logs on our console.

src/forms/FormFieldAddress/AddressDrawer/AddressDrawer.tsx
```diff
@@ -288,7 +294,7 @@ const AddressDrawer = (props: AddressDrawerProps): ReactElement => {
 					label: "States",
 					size: "sm",
 					inputSettings: {
-						options: listOfStates,
+						getOptions: () => getOptionsStates(state.data.country?.value),
```
JAndresHJGlobant, line 297: Will ``state.data.country`` have always a value? If not we should use an optional chaining operator to avoid an error.

```
getOptions: () => getOptionsStates(state.data?.country?.value)
```

JAndresHJGlobant (changes requested):
I try adding, editing, and removing Addresses in Storybook and it's working perfectly fine, but its corresponding unit tests are failing, my guess would be that the buttons to be rendered in the Drawer are commented.

src/forms/FormFieldAddress/AddressDrawer/AddressDrawer.tsx
mauriciodeleonc: Data will always exist, data.country might not, but it already has an optional chaining operator

PR #301, opened 2022-11-30
MOS-885: Update grid view to latest design

JAndresHJGlobant:
## What's included?

- Styles updated according to Figma's specs
- Bulk actions and "Select All" are moved to the top of the Grid 
- The “Select all” checkbox has an intermediate state when selecting 1 or more cards. 
- When selecting a card, makes all checkboxes visible


src/components/DataView/DataViewActionsRow/DataViewActionsRow.styled.ts
```diff
@@ -5,3 +5,13 @@ export const DataViewActionsRowWrapper = styled.div`
...
 	justify-content: space-between;
 `;
+
+export const RightControlsContainer = styled.div`
+	.custom-checkbox {
+		margin-right: 6px;
+	}
+`;
+
+export const LeftControlsContainer = styled.div`
```
mauriciodeleonc, line 15: I feel the names are swapped, the LeftControlsContainer has things that render on the right, and the RightControlsContainer has things that render on the left. 

Also, please add a margin-left auto because if  devs don't pass bulk actions then all the buttons get pulled to the left.

src/components/DataView/DataViewActionsRow/DataViewActionsRow.styled.ts
```diff
@@ -5,3 +5,13 @@ export const DataViewActionsRowWrapper = styled.div`
 	display: flex;
 	justify-content: space-between;
 `;
+
+export const RightControlsContainer = styled.div`
+	.custom-checkbox {
+		margin-right: 6px;
```
mauriciodeleonc, line 11: I think this might be 14px because it seems the space between the checkbox and the bulk button should be 24px and the checkbox already has 10

src/components/DataView/DataViewDisplayGrid/DataViewDisplayGrid.styled.ts
```diff
@@ -0,0 +1,139 @@
...
+		max-width: 100%;
+		min-width: 100%;
+	}
+
+	& > .grid > .cell h2 {
+		color: ${theme.newColors.almostBlack["100"]};
+		margin: 8px 0 8px 0;
+		max-width: 161px;
+		font-size: 13px;
+		font-weight: 590;
```
mauriciodeleonc, line 58: I know figma has 590 but let's use our semibold which is 600, the 590 doesn't exist in the "official" weights. I'll talk with Fa about this

src/components/DataView/DataViewDisplayGrid/DataViewDisplayGrid.styled.ts
```diff
@@ -0,0 +1,139 @@
...
+		white-space: nowrap;
+		text-overflow: ellipsis;
+		overflow: hidden;
+	}
+
+	& > .grid > .cell h3 {
+		margin: 0;
+		font-weight: 400;
```
mauriciodeleonc, line 66: Please update this to use our theme.fontWeight.normal. That way if it changes in any way in the future we won't need to update it manually.

mauriciodeleonc (approved):

src/components/DataView/DataViewActionsRow/DataViewActionsRow.styled.ts
JAndresHJGlobant: Actually, the space should be 20px and this needs to change to 4px because as you said we have the following:

10px checkbox + 4px extra + 2px margin of the bulk button + 4px padding of the bulk button = 20px

image

mauriciodeleonc (approved):

PR #307, opened 2022-12-05
MOS-883: Advanced Selection Selection Limits

cristianvj23:
# What's include?
- Added selectLimit prop to FormFieldAdvancedselection component
- Updated form docs
- Added a number knob to pass selectLimit


src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -323,7 +337,7 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 				{
 					name: "checkboxList",
 					type: "checkbox",
-					disabled: fieldDef?.disabled,
+					disabled: checkboxListDisabled,
```
mauriciodeleonc, line 340: We still want to disable the field when fieldDef.disabled === true or when checkboxListDisabled === true.

```suggestion
					disabled: fieldDef?.disabled || checkboxListDisabled,

```

src/forms/FormFieldAdvancedSelection/AdvancedSelection.stories.tsx
```diff
@@ -30,6 +30,7 @@ export const Playground = (): ReactElement => {
 	);
 	const getOptionsLimit = text("Get options limit", "5");
 	const createNewOptionsKnob = boolean("Create new option", true);
+	const selectLimit = number("Select limit", 5);
```
mauriciodeleonc, line 33: We still want devs / users to be able to fully play around with the component in the playground. If we keep this knob as number, they won't be able to test how to select all options (unless they know the max number of options and put that number here). I think we should keep this as string and parse it to number in line 98. I'll add an example there.

src/forms/FormFieldAdvancedSelection/AdvancedSelection.stories.tsx
```diff
@@ -93,7 +94,8 @@ export const Playground = (): ReactElement => {
 						optionsOrigin === "DB" && getOptionsLimit
 							? getOptionsLimit
 							: undefined,
-						createNewOption: createNewOptionsKnob ? createNewOption : undefined
+						createNewOption: createNewOptionsKnob ? createNewOption : undefined,
+						selectLimit
```
mauriciodeleonc, line 98: Example:
```suggestion
						selectLimit: selectLimit.trim() !== "" || !isNaN(Number(selectLimit)) ? Number(selectLimit) : undefined

```

src/forms/FormFieldAdvancedSelection/AdvancedSelection.stories.tsx
```diff
@@ -206,6 +209,17 @@ export const KitchenSink = (): ReactElement => {
 						createNewOption
 					}
 				},
+
+				{
+					name: "selectLimitOfOptions",
+					label: "Advanced selection with selectLimit prop",
```
mauriciodeleonc, line 215: I'd suggest adding this at the end so users know what to expect.
```suggestion
					label: "Advanced selection with selectLimit prop (Max 2 options)",

```

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -43,6 +43,7 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 	const [filteredOptions, setFilteredOptions] = useState<MosaicLabelValue[]>([]);
 	const [canLoadMore, setCanLoadMore] = useState<boolean>(true);
 	const [filter, setFilter] = useState({ prev: "options", new: "options" });
+	const [checkboxListDisabled, setCheckboxListDisabled] = useState(fieldDef.disabled)
```
mauriciodeleonc, line 46: Please don't forget the ";" at the end of each line. It might not be as important in javascript but in typescript it can lead to production errors.

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -64,6 +65,9 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 	useEffect(() => {
 		let isMounted = true;
 		if (value?.length > 0 && isModalOpen) {
+			if (value.length >= fieldDef.inputSettings?.selectLimit) {
+				setCheckboxListDisabled(true)
```
mauriciodeleonc, line 69: Please don't forget the ";" at the end of each line. It might not be as important in javascript but in typescript it can lead to production errors.

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -105,6 +109,10 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 			setInternalOptions();
 		}
 
+		if (fieldDef.inputSettings?.selectLimit === 0) {
+			setCheckboxListDisabled(true)
```
mauriciodeleonc, line 113: Please don't forget the ";" at the end of each line. It might not be as important in javascript but in typescript it can lead to production errors.

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -261,6 +269,9 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 	}, [fieldDef?.disabled, options]);
 
 	const deleteSelectedOption = (newOptions: MosaicLabelValue[]) => {
+		if (newOptions.length <= fieldDef.inputSettings?.selectLimit) {
+			setCheckboxListDisabled(false)
```
mauriciodeleonc, line 273: Please don't forget the ";" at the end of each line. It might not be as important in javascript but in typescript it can lead to production errors.

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -281,6 +292,9 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 	 * @param checkedOptions
 	 */
 	const checkboxListChanged = (checkedOptions: MosaicLabelValue[]) => {
+		if (checkedOptions.length >= fieldDef.inputSettings?.selectLimit) {
+			setCheckboxListDisabled(true)
```
mauriciodeleonc, line 296: Please don't forget the ";" at the end of each line. It might not be as important in javascript but in typescript it can lead to production errors.

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -261,6 +269,9 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 	}, [fieldDef?.disabled, options]);
 
 	const deleteSelectedOption = (newOptions: MosaicLabelValue[]) => {
+		if (newOptions.length <= fieldDef.inputSettings?.selectLimit) {
```
mauriciodeleonc, line 272: I think the condition should only be "less than" because if we keep the "equal" we would run into an issue with the code on line 68 `value.length >= fieldDef.inputSettings?.selectLimit ? setCheckboxListDisabled(true)` because we're saying: "When the amount of options is equal than or bigger than the limit then we disable the list" but we're also saying "When the amount of options is equal than or less than the limit then we enable the list".

```suggestion
		if (newOptions.length < fieldDef.inputSettings?.selectLimit) {

```

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -64,6 +65,9 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 	useEffect(() => {
 		let isMounted = true;
 		if (value?.length > 0 && isModalOpen) {
+			if (value.length >= fieldDef.inputSettings?.selectLimit) {
```
mauriciodeleonc, line 68: I would strongly suggest moving all of these to a function and then just calling it when needed. For example:
```
// my checkboxList disabler function
const disableCheckboxList = useCallback((optionsToCompare) => {
    const selectLimit =  fieldDef?.inputSettings?.selectLimit;  

   if (!selectLimit || selectLimit < 0) return;

    const disableList = selectLimit === 0 || optionsToCompare.length >= selectLimit;
    setCheckboxListDisabled(disableList);

}, [fieldDef?.inputSettings?.selectLimit]);

//Here in this useEffect
useEffect(() => {
    let isMounted = true;
    if (value?.length > 0 && isModalOpen) {
         disableCheckboxList(value);
         ...
    }
}
```

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -105,6 +109,10 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 			setInternalOptions();
 		}
 
+		if (fieldDef.inputSettings?.selectLimit === 0) {
```
mauriciodeleonc, line 112: I think this might not be needed if we update the function from above to re-execute with a useCallback with a dependency of selectLimit

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #308, opened 2022-12-05
MOS-876: Form sections pickers scroll behaviour fixed

JAndresHJGlobant:
## What's included?
Changes scroll behavior for the Form sections.

## How to test it?
I've defined sections of different sizes and tested the section highlight when clicking or scrolling. Example of mock sections

```
	const sections = useMemo(() => [
			{
				title: "Section 1",
				description: "Description for section 1",
				fields: [
					// row 1
					[["text1"], ["text2"], ["text3"]],
					// row 2
					[["check"], ["text4"], ["color"]],
					[[]],
					// row 3
					[["toggleSwitch"], ["imageUpload"]]
				]
			},
			{
				title: "Section 4",
				description: "Description for section 3",
				fields: [
					// row 1
					[[], [], []],
					// row 2
					[[], [], []],
					[[]],
					// row 3
					[[], []]
				]
			},
			{
				title: "Section 2",
				description: "Description for section 2",
				fields: [
					// row 1
					[["check"], ["toggleSwitch"], ["color"]],
					// row 2
					[[], [], []],
					// row 3
					[[]],
					// row 4
					[[], ["textEditor"]]
				]
			},
			{
				title: "Section 3",
				description: "Description for section 3",
				fields: [
					// row 1
					[[], [], []],
					// row 2
					[[], [], []],
					[[]],
					// row 3
					[[], []]
				]
			},
			{
				title: "Section 12",
				description: "Description for section 1",
				fields: [
					// row 1
					[["text1"], ["text2"], ["text3"]],
					// row 2
					[["check"], ["text4"], ["color"]],
					[[]],
					// row 3
					[["toggleSwitch"], ["imageUpload"]]
				]
			},
			{
				title: "Section 5",
				description: "Description for section 3",
				fields: [
					// row 1
					[[], [], []],
					// row 2
					[[], [], []],
					[[]],
					// row 3
					[[], []]
				]
			},
		], [fields]);
```

### DIFF
--- a/src/forms/FormNav/FormNav.tsx
+++ b/src/forms/FormNav/FormNav.tsx
@@ -73,42 +73,26 @@ const FormNav = (props: FormNavProps): ReactElement => {
 	 */
 	const handleClick = (e: MouseEvent, idx: number) => {
 		e.preventDefault();
-		const sectionId = sectionsRefs[idx]?.getAttribute("id");
-		sectionsRefs[idx].scrollIntoView({ behavior: "auto", block: "center", inline: "center"});
-		setSelectedTab(Number(sectionId));
+		sectionsRefs[idx].scrollIntoView({ behavior: "smooth", block: "start" });
 	};
 
 	useEffect(() => {
-		const navHighlighter = () => {
-			const scrollY = contentRef.current.scrollTop;
-			if (sectionsRefs.length === 0) {
-				return;
-			}
+		const observer = new IntersectionObserver((entries) => {
+			entries.forEach(entry => {
+				const sectionId = entry?.target.getAttribute("id");
 
-			sectionsRefs?.forEach(current => {
-				const currentSection = current as HTMLDivElement;
-				const sectionHeight = currentSection?.offsetHeight;
-				const sectionTop = currentSection?.offsetTop - 350;
-				const sectionId = currentSection?.getAttribute("id");
-
-				if (
-					scrollY > sectionTop &&
-					scrollY <= sectionTop + sectionHeight
-				) {
+				if (entry.isIntersecting) {
 					setSelectedTab(Number(sectionId));
 				}
-			});
-		};
+			})
+		}, { threshold: 0.9 });
 
-		const navHighlighterDebounced = debounce(navHighlighter, 200, { maxWait: 100 });
+		sectionsRefs?.forEach(section => {
+			observer.observe(section);
+		})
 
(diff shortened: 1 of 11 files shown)


src/forms/FormNav/FormNav.tsx
```diff
@@ -71,9 +71,11 @@ const FormNav = (props: FormNavProps): ReactElement => {
 	 * @param idx
 	 * @param sectionId
 	 */
-	const handleClick = (e, idx: number) => {
+	const handleClick = (e: MouseEvent, idx: number) => {
 		e.preventDefault();
-		sectionsRefs[idx].scrollIntoView({ behavior: "smooth", block: "start", inline: "nearest"});
+		const sectionId = sectionsRefs[idx]?.getAttribute("id");
+		sectionsRefs[idx].scrollIntoView({ behavior: "auto", block: "center", inline: "center"});
```
mauriciodeleonc, line 77: I have several comments here:
1. We still want to keep the behavior as smooth because we want the page to still feel responsive.
2. I believe we can completely remove the inline property, it only affects the horizontal alignment (https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#parameters)
3. I feel the block property (which affects the vertical alignment) needs to be on start. Testing around with the sections example you provided, when clicking on the tabs it takes you to the center of the section, which we don't want because there are forms that might be hundreds of fields long, so it would take them to the middle of the form.

Additionally, testing the sections you provided it seems this doesn't fully solve the issue because if you start clicking one by one, when you click on the last tab it bounces and selects the previous section (section 12).

I'm not sure if this is the best approach but I think we could remove the scroll listener when clicking, and re-add it once the "scrollIntoView"  and the setSelectedTab have finished.

src/forms/FormNav/FormNav.tsx
```diff
@@ -88,6 +90,7 @@ const FormNav = (props: FormNavProps): ReactElement => {
 				const sectionHeight = currentSection?.offsetHeight;
 				const sectionTop = currentSection?.offsetTop - 350;
 				const sectionId = currentSection?.getAttribute("id");
+
```
mauriciodeleonc, line 93: Found this other approach using observers, we might need to rethink / refactor our use effect. It's compatible with all browsers and I think might be a more "elegant" solution than the one I suggested above, but we'll have to give them a try.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #315, opened 2022-12-12
MOS-882: Advanced selection oddities issues

JAndresHJGlobant:
## What's included?

Fix issues with the selected options


src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -77,19 +103,8 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
...
-				);
-
-				dispatch(
-					formActions.setFieldValue({
-						name: "listOfChips",
-						value: value
-					})
-				);
+				dispatchChipList(value);
+				dispatchCheckboxList(value);
```
mauriciodeleonc, line 107: We don't want to dispatchCheckboxList here because we already have a useEffect in place that dispatches the checkbox list every time the chipList changes. That way we ensure the "master" of the data is chipList.

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -274,46 +289,33 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
...
-			})
-		);
-
-		dispatch(
-			formActions.setFieldValue({
-				name: "checkboxList",
-				value: newOptions,
-			})
-		);
+		dispatchCheckboxList(newOptions);
```
mauriciodeleonc, line 292: Same comment as above

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -348,8 +350,9 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 					size: "100%",
 					onChangeCb: checkboxListChanged,
 					inputSettings: {
-						options: fieldDef?.inputSettings?.options && filteredList,
-						getOptions: fieldDef?.inputSettings?.getOptions && (() => filteredList),
+						options:
+							(fieldDef?.inputSettings?.options && filteredList) ||
+							(fieldDef?.inputSettings?.getOptions && (() => filteredList)),
```
mauriciodeleonc, line 355: I'm not sure why I can't comment on that line, but I believe we can delete the getMoreOptions from the dependency array, it's not being used here and when doing the research / testing if I remember correctly it caused re-renders on the fields.

mauriciodeleonc (changes requested):

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
JAndresHJGlobant: If we removed ``getMoreOptions`` as a dependency we lost the ``selectLimit`` functionality. I tried to put ``fieldDef?.inputSettings?.selectLimit`` as a dependency but it did not work

automation_testing/tests/Components/form/validators.spec.ts
```diff
@@ -64,8 +64,7 @@ test.describe.parallel("Components - Form - Validators", () => {
 		await validatorPage.title.click();
 		const error = validatorPage.error;
 		expect(await error.textContent()).toBe(validatorData.emailError);
-		await validatorPage.emailField.selectText();
-		await validatorPage.clearAllValuesFromField();
+		await validatorPage.clearAllValuesFromField(validatorPage.emailField);
```
mauriciodeleonc, line 67: I can't add the comment where it belongs but I got this error when running the tests
image

mauriciodeleonc (changes requested):

automation_testing/tests/Components/form/validators.spec.ts
torres-nahuel: I wasn't able to replicate the error but I know what is causing it. I implemented a fix, please run the test again and tell me if it passes.

mauriciodeleonc (approved):

PR #316, opened 2022-12-13
MOS-900: Form disable logic on load not disabling correctly

JAndresHJGlobant:
## What's included?

Disabling form logic changed when it is prepopulated.

## How to test it?

Go to the Form playground story and prepopulate it. You should see the screen block for a second and then all the fields filled.

### DIFF
--- a/src/components/Form/Form.test.tsx
+++ b/src/components/Form/Form.test.tsx
@@ -446,7 +446,7 @@ describe("REDUCERS: FORM_RESET", () => {
 					validating: {},
 					custom: {},
 					validForm: false,
-					disabled: null,
+					disabled: false,
 				},
 			},
 		}
@@ -857,10 +857,6 @@ describe("DISPATCHERS: setFormValues", () => {
 				}],
 				calls: [
 					{
-						type: "FORM_START_DISABLE",
-						value: true,
-					},
-					{
 						type: "FIELD_ON_CHANGE",
 						name: "field1",
 						value: "value1",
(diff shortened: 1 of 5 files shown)


src/components/Form/formActions.ts
```diff
@@ -181,13 +176,16 @@ export const formActions = {
 					})
 				);
 			}
-
+		}
+	},
+	disableForm(disabled: boolean) {
```
mauriciodeleonc, line 181: Let's keep the args as object to maintain the same "structure" as the other actions.
```suggestion
	disableForm({disabled = false}: {disabled: boolean}) {

```

src/components/Form/Form.test.tsx
```diff
@@ -870,6 +866,35 @@ describe("DISPATCHERS: setFormValues", () => {
...
+						value: true,
+					}
+				]
+			}
+		},
+		{
+			name: "Enables the form",
+			args: {
+				action: "disableForm",
+				args: [false],
```
mauriciodeleonc, line 896: Please update to object when finishing updating the action's args.

src/components/Form/Form.test.tsx
```diff
@@ -870,6 +866,35 @@ describe("DISPATCHERS: setFormValues", () => {
...
+	runTests(tests, "dispatch");
+});
+
+describe("DISPATCHERS: disableForm", () => {
+	const tests = [
+		{
+			name: "Disables the form",
+			args: {
+				action: "disableForm",
+				args: [true],
```
mauriciodeleonc, line 883: Please update to object when finishing updating the action's args.

src/components/Form/Form.tsx
```diff
@@ -68,6 +68,7 @@ const Form = (props: FormProps) => {
 			let values: MosaicObject;
 
 			if (getFormValues) {
+				await dispatch(formActions.disableForm(true));
```
mauriciodeleonc, line 71: I think we might want to disable the form before this if statement, this way it can be disabled even while assigning values through the defaultValue prop, then it'll get enabled back again at the end.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #317, opened 2022-12-13
MOS-877: Update indentation of fields

cristianvj23:
# What's include?
- Updated indentation of fields

### DIFF
--- a/src/components/DataView/DataView.tsx
+++ b/src/components/DataView/DataView.tsx
@@ -32,7 +32,7 @@ const StyledWrapper = styled.div`
 
 	& > .headerActions .grid {
 		border-bottom: 2px solid ${theme.newColors.grey2["100"]};
-		padding-bottom: 20px;
+		padding-bottom: 8px;
 	}
 
 	& > .viewContainer {
--- a/src/components/DataView/DataView.tsx
+++ b/src/components/DataView/DataView.tsx
@@ -27,7 +27,7 @@ const StyledWrapper = styled.div`
 	& > .headerActions {
 		display: flex;
 		flex-direction: column;
-		padding: 20px;
+		padding: 8px 24px;
 	}
 
 	& > .headerActions .grid {
--- a/src/forms/TopComponent/TopComponent.styled.ts
+++ b/src/forms/TopComponent/TopComponent.styled.ts
@@ -11,7 +11,7 @@ const RESPONSIVE_BREAKPOINT = BREAKPOINTS.topComponent.responsiveView + "px";
 
 export const FormContent = styled.div`
 	& .layout {
-		padding: ${pr => pr.view === "DRAWER" ? "15px 40px" : (pr.view === "RESPONSIVE" || pr.view === "DESKTOP" || pr.view === "BIG_DESKTOP") ? "24px" : "0px"};
+		padding: ${pr => pr.view === "MOBILE" ? "0px" :  "24px"};
 	}
 	overflow-y: auto;
 	width: 100%;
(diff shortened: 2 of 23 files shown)


src/forms/TopComponent/TopComponent.styled.ts
```diff
@@ -11,7 +11,7 @@ const RESPONSIVE_BREAKPOINT = BREAKPOINTS.topComponent.responsiveView + "px";
 
 export const FormContent = styled.div`
 	& .layout {
-		padding: ${pr => pr.view === "DRAWER" ? "15px 40px" : (pr.view === "RESPONSIVE" || pr.view === "DESKTOP" || pr.view === "BIG_DESKTOP") ? "15px 20px 0px 20px" : "0px"};
+		padding: ${pr => pr.view === "DRAWER" ? "15px 40px" : (pr.view === "RESPONSIVE" || pr.view === "DESKTOP" || pr.view === "BIG_DESKTOP") ? "24px" : "0px"};
```
mauriciodeleonc, line 14: The idea here is that all variants of forms will have 24px around their layout. The only variant that might not is the mobile because of the reduced screen-size. 
```suggestion
		padding: ${pr => pr.view === "MOBILE" ? "0px" :  "24px"};

```

src/forms/TopComponent/Views/DesktopView.tsx
```diff
@@ -22,7 +22,7 @@ const BIG_SCREEN_BREAKPOINT = BREAKPOINTS.topComponent.bigScreenView + "px";
 
 const DesktopViewColumn = styled(StyledColumn)`
 	justify-content: space-between;
-	padding: ${pr => pr.sections ? "24px 20px 0px 20px" : "24px 20px 20px 20px"};
+	padding: ${pr => pr.sections ? "24px 24px 0px 24px" : "24px 24px 16px 24px"};
```
mauriciodeleonc, line 25: Now that almost all variants will share the same sizes, spaces, and styling, I'm confident we could move all of this code into "StyledColumn" and delete "DesktopViewColumn"

src/forms/TopComponent/Views/ResponsiveView.tsx
```diff
@@ -17,7 +17,7 @@ import TitleWrapper from "../Utils/TitleWrapper";
 import { BaseTopComponentProps, TopComponentProps } from "../TopComponentTypes";
 
 const ResponsiveViewColumn = styled(StyledColumn)`
-	padding: ${pr => pr.sections ? "24px 20px 0px 20px" : "24px 20px 20px 20px"};
+	padding: ${pr => pr.sections ? "24px 24px 0px 24px" : "24px 24px 16px 24px"};
```
mauriciodeleonc, line 20: Now that almost all variants will share the same sizes, spaces, and styling, I'm confident we could move all of this code into "StyledColumn" and delete "ResponsiveViewColumn"

src/components/DataView/DataView.tsx
```diff
@@ -27,7 +27,7 @@ const StyledWrapper = styled.div`
 	& > .headerActions {
 		display: flex;
 		flex-direction: column;
-		padding: 20px;
+		padding: 8px 24px;
```
mauriciodeleonc, line 30: I talked with Owen about the line below the grid actions. He said we should be changing it to whatever we feel looks best. So I think we should update .headerActions .grid padding-bottom: 8px;

mauriciodeleonc (changes requested):
Additional comments:
Overall, all spacing will now be a multiple of 8, so we should be changing it to the closest if we see one. For example, if something has a padding of 15, then it should be now be 16px, etc.

This ticket mentions that the layout area of the form should be 24px, this also includes when the fields are inside of a section, so please in Section.tsx update "StyleRows" (line 37) to say "16px 24px" at the end. Also, line 24 should be changed to "16px 24px", and line 29 to "24px 24px 0px 24px". This way everything inside of sections will be aligned.

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #319, opened 2022-12-14
MOS-897: Update buttons in the filter dropdowns

cristianvj23:
# What's include?
- Updated filterDropdown buttons


src/components/DataViewFilterDropdownButtons/DataViewFilterDropdownButtons.tsx
```diff
@@ -53,8 +52,7 @@ function DataViewFilterDropdownButtons(props: DataViewFilterDropdownButtonsProps
 				<Button label={t("mosaic:common.clear")} color="black" variant="text" onClick={props.onClear}/>
 			</ButtonRow>
 			<StyledButtonRow>
```
mauriciodeleonc, line 54: I think StyledButtonRow might not be needed anymore because we no longer have 2 buttons, we could update it to just use ButtonRow.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #321, opened 2022-12-16
MOS-655: left side form sections dont appear to be working

JAndresHJGlobant:
## What's included?

- Use ResizeObserver attached to the Form container to define the user's view
- Updates unit tests using a mock ResizeObserver
- Removes media queries and uses classNames with the view name
- Creates a context and a custom hook to get the view


src/components/Form/Form.styled.ts
```diff
@@ -3,7 +3,6 @@ import styled from "styled-components";
 export const StyledForm = styled.form`
 	display: flex;
 	flex-direction: column;
-	justify-content: space-between;
```
mauriciodeleonc, line 6: Did you delete this because of the blank space that appear between the layout and the top component? If you did then this doesn't fully solve the issue. The issue has to do more with this piece of code in Form.tsx
`useEffect(() => {
		setTopComponentHeight(topComponentRef.current?.offsetHeight);
	}, [topComponentRef.current?.offsetHeight]);`

Here the topComponent's offsetHeight is not being updated when the view changes, so it stays with the previous one, you can test this by opening the "Form with layout story" in a small screen size, then making it big until the nav links move to the left. It would look something like this: 
image

What I think might solve the issue is adding / changing the dependency of that useEffect to also include view, that way when it changes it also updates the height

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #328, opened 2022-12-26
MOS-848: Update all references to simply gray

JAndresHJGlobant:
## What's included?

Replaces simplyGray references with the new color syntax.

## Notes

simplyGrayOpacity (190, 190, 190, 0.3) was replaced with the variant using 0.2 opacity since we do not have a 0.3 variant.

### DIFF
--- a/src/components/Field/InstructionText.styled.ts
+++ b/src/components/Field/InstructionText.styled.ts
@@ -2,7 +2,7 @@ import styled from "styled-components"
 import theme from "@root/theme";
 
 export const InstructionTextWrapper = styled.div`
-  border-left: 1px solid ${theme.newColors.simplyGrey["100"]};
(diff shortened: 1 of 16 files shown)


src/components/Field/InstructionText.styled.ts
```diff
@@ -2,7 +2,7 @@ import styled from "styled-components"
 import theme from "@root/theme";
 
 export const InstructionTextWrapper = styled.div`
-  border-left: ${theme.borders.simplyGray};
+  border-left: 1px solid ${theme.newColors.simplyGrey["100"]};
```
mauriciodeleonc, line 5: We already have a border in place for this color, so we don't have to repeat it's style in different places. I'll update this directly.

mauriciodeleonc (approved):

PR #329, opened 2022-12-27
MOS-907: Mosaic saved view def has required keys that don't make sense

cristianvj23:
# What's include?
- Updated savedViewState keys as not required

### DIFF
--- a/src/components/DataView/DataViewTitleBar/DataViewTitleBarTypes.ts
+++ b/src/components/DataView/DataViewTitleBar/DataViewTitleBarTypes.ts
@@ -1,13 +1,4 @@
-import { DataViewProps } from "../DataViewTypes";
-
-interface SavedViewState {
-	limit?: DataViewProps["limit"];
-	sort?: DataViewProps["sort"];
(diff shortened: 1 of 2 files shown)


src/components/DataView/DataViewTitleBar/DataViewTitleBarTypes.ts
```diff
@@ -1,13 +1,12 @@
...
+import { DataViewProps } from "../DataViewTypes";
 
 interface SavedViewState {
-	limit: number;
-	sort: DataViewSort;
-	display: string;
-	filter: MosaicObject;
-	activeFilters: string[];
-	activeColumns: string[];
+	limit?: DataViewProps["limit"];
```
mauriciodeleonc, line 4: Looking at the code and the commits, it seems this interface was recently created and there was another one previously created in DataViewTypes.ts so in order to not have duplicates I deleted this one and imported the other directly here.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #330, opened 2022-12-27
MOS-901: Browse saved views has bad spacing

cristianvj23:
# What's include?
- Updated saved view drawer

### DIFF
--- a/src/components/DataView/DataView.tsx
+++ b/src/components/DataView/DataView.tsx
@@ -401,47 +401,53 @@ function DataView (props: DataViewProps): ReactElement  {
 			${props.sticky ? "sticky" : ""}
 		`}
 		>
-			<div className="headerRow title">
-				<DataViewTitleBar
-					title={props.title}
-					buttons={props.buttons}
-					savedViewEnabled={savedViewEnabled}
-					savedView={props.savedView}
-					savedViewState={savedViewState}
-					savedViewCallbacks={savedViewCallbacks}
-					savedViewAllowSharedViewSave={(props.savedViewAllowSharedViewSave !== undefined) ? props.savedViewAllowSharedViewSave : false }
-					loading={props.loading}
-					filter={props.filter}
-					filters={props.filters}
-					activeFilters={props.activeFilters}
-					onActiveFiltersChange={props.onActiveFiltersChange}
-				/>
-			</div>
-			<div className="headerActions">
-				<DataViewActionsRow
-					activeColumnObjs={activeColumnObjs}
-					columns={props.columns}
-					bulkActions={props.bulkActions}
-					checked={state.checked}
-					display={display}
-					displayControlEnabled={displayControlEnabled}
-					displayOptionsFull={displayOptionsFull}
-					limit={props.limit}
-					limitOptions={props.limitOptions}
-					onLimitChange={props.onLimitChange}
-					onDisplayChange={props.onDisplayChange}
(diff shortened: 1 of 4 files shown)


src/components/DataView/DataView.tsx
```diff
@@ -401,47 +401,53 @@ function DataView (props: DataViewProps): ReactElement  {
...
-					onColumnsChange={props.onColumnsChange}
-					onCheckAllClick={onCheckAllClick}
-					onSortChange={props.onSortChange}
-					sort={props.sort}
-					data={props.data}
-					checkedAllPages={state.checkedAllPages}
-				/>
-			</div>
+			{
+				props.title &&
```
mauriciodeleonc, line 405: The DataViewTitleBar doesn't depend 100% on the title. The way this condition is written says "If there's no title then we can't render the component" but DataViews can render the title bar with just buttons, with just active filters, etc, etc (any of the props it receives). What you'd need to do here is evaluate every single prop and not render the component **only** when none are present, if at least 1 prop gets passed then we render it.

src/components/DataView/DataView.tsx
```diff
@@ -401,47 +401,53 @@ function DataView (props: DataViewProps): ReactElement  {
...
+							loading={props.loading}
+							filter={props.filter}
+							filters={props.filters}
+							activeFilters={props.activeFilters}
+							onActiveFiltersChange={props.onActiveFiltersChange}
+						/>
+					</div>
+			}
+			{
+				props.bulkActions &&
```
mauriciodeleonc, line 424: The DataViewActionsRow doesn't depend 100% on the bulkActions. DataViews can render the actions row with just display, control, limit, etc, etc (any of the props it receives). What you'd need to do here is evaluate every single prop and not render the component **only** when none are present, if at least 1 prop gets passed then we render it.

mauriciodeleonc (changes requested):

src/components/DrawerContent/DrawerContent.tsx
```diff
@@ -11,14 +15,55 @@ const StyledDiv = styled.div`
 type DrawerContentProps = DrawerTopBarProps & DrawerBottomProps;
 
 function DrawerContent(props: DrawerContentProps) {
+
+	const { t } = useMosaicTranslation();
+
+	const [drawerButtons, setDrawerButtons] = useState<ButtonProps[]>([]);
+
+	useEffect(() => {
```
mauriciodeleonc, line 23: This is a good approach but I have 2 recommendations here:
1. Change useEffect to useMemo (https://reactjs.org/docs/hooks-reference.html#usememo). This stores the result in memory and doesn't reassign it with every render, only when those props change. The way it is now it will re-render / re-create the buttons every single time any of it's parent components re-renders.
2. I don't like depending on the state because useState is asynchronous, this means the real value won't always be the rendered value because react has to wait for the component to finish re-rendering to update the state. Using useMemo also solves this issue because it assigns the value to an actual variable.

mauriciodeleonc: Here's how the result would look like:
```suggestion
	const drawerButtons = useMemo(() => {
	        let buttons: ButtonProps[] = [];
	        ... all your code
	        return buttons;
	        }, [the deps you already have]);
```

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #345, opened 2023-01-09
MOS-845: Update all references to gray2

JAndresHJGlobant:
## What's included?

- Replaces gray200 with gray2 new color variant
- Removes grayHover color since it was the same as gray2


mauriciodeleonc (approved):

mauriciodeleonc (changes requested):
Locators seem to be moved
image

mauriciodeleonc (approved):

PR #350, opened 2023-01-11
Mos 931: Update Advanced Selection's checkboxList height

cristianvj23:
# What's include?
- Added some component refs to AdvancedSelectionDrawer component to calculate CheckboxList height


src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -43,12 +43,38 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
...
+	const [formLayoutPadding, setFormLayoutPadding] = useState("0px");
 
 	const chipListRef:{ current?: HTMLDivElement } = useRef();
+	const loadMoreButtonRef:{ current?: HTMLDivElement } = useRef();
+	const searchInputRef:{ current?: HTMLDivElement } = useRef();
 
 	const { state, dispatch } = useForm();
 
-	const chipListHeight: number = chipListRef?.current?.offsetHeight ? chipListRef?.current?.offsetHeight + 30 : 0;
+	const chipListHeight: string = chipListRef?.current?.offsetHeight
```
mauriciodeleonc, line 55: This is a good approach, I just have some observations to optimize memory and prevent re-renders:
1. The value could be stored using a useMemo with the ref as dependency, that way it won't re-calculate with every other change to the component.
2. Line 56 doesn't need all the question marks because you're already making sure the object exists one line above, so you could just have it like this: ${chipListRef.current.offsetHeight}

These comments go to all other variables you defined under this.

src/components/Form/FormTypes.tsx
```diff
@@ -26,4 +26,6 @@ export interface FormProps {
 	buttons?: ButtonProps[];
 	tooltipInfo?: string;
 	showActive?: boolean;
+	setTopComponentDrawerHeight?: Dispatch<SetStateAction<string>>;
```
mauriciodeleonc, line 29: I don't like the idea of passing these refs / setStates to the Form because it's not its responsibility and also this can cause issues with other developers passing refs. I'm thinking on another solution that might involve context / reducers, that way the only components responsible are the ones involved (AdvancedSelection, TopComponent and FormLayout)

src/components/Form/Form.tsx
```diff
@@ -46,6 +48,14 @@ const Form = (props: FormProps) => {
 
 	useEffect(() => {
 		setTopComponentHeight(topComponentRef.current?.offsetHeight);
+		setTopComponentDrawerHeight && setTopComponentDrawerHeight(`${topComponentRef.current?.offsetHeight}px`);
```
mauriciodeleonc, line 51: Several comments here:
1. Same comment as in FormTypes.tsx
2. This wouldn't always work as expected because formContentRef wasn't added to the dependency array. 

Once the context / reducer approach gets solved I believe all of this can be deleted and moved to where it's actually needed.

src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx
```diff
@@ -394,6 +421,8 @@ const AdvancedSelectionDrawer = (props: AdvanceSelectionDrawerPropTypes): ReactE
 				fields={fields}
 				dialogOpen={dialogOpen}
 				handleDialogClose={handleDialogClose}
+				setTopComponentDrawerHeight={setTopComponentDrawerHeight}
```
mauriciodeleonc, line 424: Same comment as FormTypes.ts

mauriciodeleonc (changes requested):

src/components/Form/Form.tsx
```diff
@@ -48,13 +50,23 @@ const Form = (props: FormProps) => {
...
 
-		if (formContentRef?.current?.children[0] && setFormLayoutPadding) {
-			const layoutPaddingTop = window?.getComputedStyle(formContentRef.current.children[0]).getPropertyValue("padding-top");
-			const layoutPaddingBottom = window?.getComputedStyle(formContentRef.current.children[0]).getPropertyValue("padding-bottom");
-			setFormLayoutPadding(`${layoutPaddingTop} - ${layoutPaddingBottom}`);
-		}
+		dispatchRef && topComponentRef.current && dispatchRef({
+			type: "update",
+			ref: {
+				id: "topComponentDrawerRef",
```
mauriciodeleonc, line 57: Same comment as RefsReducer:
```suggestion
				topComponentDrawerRef: topComponentRef.current
```

src/components/Form/Form.tsx
```diff
@@ -48,13 +50,23 @@ const Form = (props: FormProps) => {
...
+				current: topComponentRef.current
+			}
+		});
+
+		formContentRef?.current?.children[0] && dispatchRef &&
```
mauriciodeleonc, line 62: This wouldn't always work as expected because formContentRef it's not in the dependency array, it should be but **not** in this useEffect, it should be in its own because then it will trigger changes in the other dispatchRef

src/components/Form/Form.tsx
```diff
@@ -48,13 +50,23 @@ const Form = (props: FormProps) => {
...
+			dispatchRef({
+				type: "update",
+				ref: {
+					id: "formLayoutRef",
```
mauriciodeleonc, line 66: Same comment as RefsReducer:
```suggestion
					formLayoutRef: formContentRef.current.children[0]

```

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #353, opened 2023-01-18
MOS-896: Add google address lookup mechanics to Address field

JAndresHJGlobant:
## What's included?

- U mechanics so that it drops down a list of options that closely match. If I select one, it should fill in the address, city, state, zip and country
- When Google's Autocomplete API does not find a value for either address, city, country, or postal code a snackbar is rendered telling the user that those fields need to be filled manually.
- Refactor of the LocationSearchInput component to be reusable by the Map and the Address components.
- Updates unit tests


src/forms/FormFieldAddress/AddressAutocomplete/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from "./AddressAutocomplete";
+export * from "./AddressAutocomplete";
```
mauriciodeleonc, line 2: We should also be exporting the types.

src/forms/FormFieldAddress/AddressAutocomplete/AddressAutocompleteTypes.ts
```diff
@@ -0,0 +1,21 @@
...
+	value: string;
+	onChange: (address: string) => void;
+	className?: string;
+	onSelect: (value: any) => Promise<void>;
+	textField?: TextFieldProps;
+	fieldSize?: string;
+	placeholder?: string
+}
+
+export interface InputSettings {
```
mauriciodeleonc, line 13: This doesn't seem to be used anywhere else in the project, also... it's the same interface as AddressAutocompleteProps, so we should just have 1

src/forms/FormFieldAddress/AddressAutocomplete/AddressAutocomplete.styled.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const SuggestionsContainer = styled.div`
+  border: ${theme.borders.fieldGray};
+  font-family: ${theme.fontFamily};
+`;
+
+export const SuggestionsDescriptionContainer = styled.div`
+  background-color: ${({ isSuggestionActive }) => isSuggestionActive ? theme.colors.gray200 : "white"};
+  cursor: pointer;
+	width: ${({ fieldSize }) => fieldSize};
```
mauriciodeleonc, line 36: We need to remove 2px from the fieldSize because of the 1px border, else the container gets on top of the border, this can be solved if we move some of the logic between components, for example: Here we can keep width as 100%, and move the padding from SuggestionDescription up to this component. That way the one applying the padding is the parent and not the child. Also this has some extra indentation

src/forms/FormFieldAddress/AddressAutocomplete/AddressAutocomplete.styled.tsx
```diff
@@ -0,0 +1,48 @@
...
+    display: flex;
+  }
+`;
+
+export const StyledClearIcon = styled(ClearIcon)`
+  cursor: pointer;
+  color: ${theme.newColors.almostBlack["100"]};
+`;
+
+export const SuggestionsContainer = styled.div`
```
mauriciodeleonc, line 28: We should change the position of this component to absolute, that way the dropdown won't move all other elements around.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #357, opened 2023-01-23
MOS-941: No results message misaligned

JAndresHJGlobant:
## What's included?
Adds margin to the "No results" message to be aligned to the Data View table.

### DIFF
--- a/src/components/DataView/DataView.tsx
+++ b/src/components/DataView/DataView.tsx
@@ -17,7 +17,7 @@ const StyledWrapper = styled.div`
 	flex-direction: column;
 
 	& > .noResults {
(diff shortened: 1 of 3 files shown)


src/components/DataView/DataView.tsx
```diff
@@ -16,6 +16,10 @@ const StyledWrapper = styled.div`
 	display: flex;
 	flex-direction: column;
 
+	& > .noResults {
+		margin-left: 20px;
```
mauriciodeleonc: Let's keep consistency between both sides. Please change this to have 20px both on the left and right.

mauriciodeleonc (approved):

PR #360, opened 2023-01-24
MOS-940: Merge text area with text field

JAndresHJGlobant:
## What's included?

- Merge text area with text field
- Remove all references to textArea
- Update docs
- Adds minRows and maxRows props to text field.


src/components/Form/Form.stories.mdx
```diff
@@ -774,6 +774,8 @@ This field implements the [**FieldDef**](#generic-field-props-fielddef) interfac
 | **`placeholder`** | `string` | optional - The hint displayed in the input before the user enters a value. |
 | **`prefixElement`** | `JSX.Element` | optional - Icon at the beginning of the text field. |
 | **`type`** | `string` | optional - Type of the input element. It should be a [valid HTML5 input type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Form_%3Cinput%3E_types). |
+| **`maxRows`** | `string `\|` number` | optional - Maximum number of rows to display when multiline options is set to true. |
```
mauriciodeleonc, line 777: Please add these props to the example below

src/components/Form/Form.stories.mdx
```diff
@@ -322,7 +322,7 @@ const fields = useMemo(
 * **instructionText** - `string` optional - Instructions about how to fill the current field.
 * **disabled** - `boolean` optional - Flag that indicates whether the field can be filled or readonly.
 * **inputSettings** - `object` optional - Specific props for each specific field (Please see the Props section of each field).
-* **maxCharacters** - `number` optional - When defined, displays the max amount of characters allowed for this field next to the label (only applies for Text and TextArea).
+* **maxCharacters** - `number` optional - When defined, displays the max amount of characters allowed for this field next to the label (only applies for Text field).
```
mauriciodeleonc, line 325:
```suggestion
* **maxCharacters** - `number` optional - When defined, displays the max amount of characters allowed for this field next to the label (only applies for FormFieldText).
```

src/components/Form/Form.stories.tsx
```diff
@@ -426,7 +417,7 @@ export const Playground = (): ReactElement => {
 			description: text("Description for section 1", "Description for section 1"),
 			fields: [
 				// row 1
-				[["textField"], ["textArea"], ["check"]],
+				[["textField"], ["textField"], ["check"]],
```
mauriciodeleonc, line 420: I think it might be better to just delete it because then when the user types on any of them the value will apply to both and it might look as if there's an issue
```suggestion
				[["textField"], ["check"]],

```

src/components/Form/Form.stories.mdx
mauriciodeleonc, line 777: I know MUI's documentation says it can be string or number, but I believe we should keep it 100% as number (for both min and max).
```suggestion
| **`maxRows`** | `string` | optional - Maximum number of rows to display when multiline options is set to true. |
```

src/forms/FormFieldText/FormFieldTextTypes.tsx
```diff
@@ -25,4 +25,14 @@ export type TextFieldDef = {
 	 * input type.
 	 */
 	type?: string;
+	/**
+	 * Minimum number of rows to display when multiline option
+	 * is set to true.
+	 */
+	minRows?: string | number;
```
mauriciodeleonc, line 32: I know MUI's documentation says it can be string or number, but I believe we should keep it 100% as number (for both min and max).
```suggestion
	minRows?: number;

```

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #364, opened 2023-01-30
MOS-938: No verification of defaultValues pathway in storybook and alterations

cristianvj23:
# What's include?
- Updated form's logic to switch between defaultValues and getFormValues()


src/forms/FormFieldTextArea/FormFieldTextArea.tsx
```diff
@@ -22,7 +22,7 @@ const TextArea = (
 	return (
 		<StyledTextArea
 			id={fieldDef.name}
-			value={value}
+			value={value ?? ""}
```
mauriciodeleonc, line 25: It seems this isn't using the latest code because we deleted all of this component last week

src/components/Form/Form.stories.tsx
```diff
@@ -494,7 +636,7 @@ export const Playground = (): ReactElement => {
 					state={state}
 					fields={fields}
 					dispatch={dispatch}
-					getFormValues={loadReady && getFormValues}
+					getFormValues={showGetFormValues === "None" ? loadReady && undefined : loadReady && getFormValues}
```
mauriciodeleonc, line 639: If showGetFormValues is "None" that means getFormValues is undefined, so we don't actually care if loadReady is true or not. I'd also suggest adding the parentheses to make it easier to read.
```suggestion
					getFormValues={showGetFormValues === "None" ? undefined : (loadReady && getFormValues)}

```

src/components/Form/Form.tsx
```diff
@@ -95,9 +95,9 @@ const Form = (props: FormProps) => {
 			let values: MosaicObject;
 			await dispatch(formActions.disableForm({ disabled: true }));
 
-			if (getFormValues) {
-				values = await getFormValues();
-			} else {
+			values = getFormValues ? await getFormValues() : undefined;
```
mauriciodeleonc, line 98: Great and simple approach! 👍

src/components/Form/Form.stories.tsx
```diff
@@ -79,6 +79,8 @@ export const Playground = (): ReactElement => {
 
 	const showState = boolean("Show state", false);
 	const prepopulate = boolean("Prepopulate", false);
+	const defaultValuesKnob = select("Default Values", ["None", "Has Defaults"], "None");
+	const showGetFormValues = select("GetFormValues", ["None", "Returns Undefined", "Returns Data"], "Returns Data");
```
mauriciodeleonc, line 83: From the ticket: "Ensure that there are a few more fields with defaultValue than their are fields filled by getFormValues. We need this to prove that defaultValue is not being used when getFormValues exists and returns data." In line 92, please delete some of the entries and also I'd suggest changing the text in some of them to include the word "getFormValues"

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #367, opened 2023-02-07
MOS-947: BulkActions shouldn't be required to render checkboxes in DataView

cristianvj23:
# What's include?
- Removed validBulkActions conditions from dataView


mauriciodeleonc (changes requested):
Please make sure all the changes are working correctly before submitting the PR.
When you disable the "bulkActions" prop and you click on any checkbox you'll see that there's some white space where the columns should be, this is because on "DataViewTHead.tsx" the DataViewBulkActionsButtonsRow (line 195) is not being rendered, which is fine, we want that to happen when no bulkActions get passed, but we should make sure that we keep the space. You can fix this by adding a colSpan property to the StyledTh from line 186, this property tells the th how many columns to take, in this case it would be whatever the length of the columns is + 2 (same as line 196, but 1 more because we no longer have bulkActions).
image

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #370, opened 2023-02-07
TAE - MOS-952 - Update visit page logic for automation

torres-nahuel:
### Jira Ticket
https://simpleviewtools.atlassian.net/browse/MOS-952

### What's included?
The objective of this ticket is to introduce enhancements to reduce code repetition, specifically the "visitPage" method in each component and form field page. The "visitPage" method implementation goes against the principles of the DRY Practice.


mauriciodeleonc (changes requested):
Everything seems fine, all tests passing except these 3 that were just recently added.
image

PR #371, opened 2023-02-07
MOS-954: Href on icon button

JAndresHJGlobant:
## What's included?

- Adds href prop to icon button variant
- Updates styles selectors that were affected because the button was transformed into an anchor tag
- Updates onClick types


src/components/Button/Button.styled.ts
```diff
@@ -23,7 +23,7 @@ export const ButtonWrapper = styled.span`
   }
 
   & > a {
-	text-transform: uppercase;
+    text-transform: uppercase;
   }
 
   &.normalButton > button {
```
mauriciodeleonc, line 29: I feel we should make the jump and update all "button" and "a" to .MuiButtonBase-root, that way we don't have to maintain 2 different styles for button links and regular buttons.

src/components/Button/Button.styled.ts
```diff
@@ -23,7 +23,7 @@ export const ButtonWrapper = styled.span`
   }
 
   & > a {
```
mauriciodeleonc, line 25: Once we update everything to be .MuiButtonBase this whole style can go away because it will depend on what variant the users decide to pass.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #372, opened 2023-02-07
MOS-955: Remove maxCharacter prop from FieldDefBase

cristianvj23:
# What's include?
- Removed extra maxCharacters fieldDef from FormFieldText component


mauriciodeleonc (changes requested):
Everything seems good, just one observation... we should also be updating the Form.stories.mdx documentation with this new functionality.

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #376, opened 2023-02-14
MOS-961: Updated to upload component

JAndresHJGlobant:
## What's included?

- Disables upload button based on loaded and pending files.
- Adds a knob to set the seconds that each file will take to load completely.

### DIFF
--- a/src/forms/FormFieldUpload/FormFieldUpload.tsx
+++ b/src/forms/FormFieldUpload/FormFieldUpload.tsx
@@ -126,14 +126,11 @@ const FormFieldUpload = (props: MosaicFieldProps<"upload", UploadFieldInputSetti
 	const handleNewFileUpload = async (e) => {
 		const newFiles: File[] = Array.from(e.target.files);
 
-		const pendingWithoutError = Object.values(pendingFiles).filter((pendingFile: {error: string}) => pendingFile.error === undefined).length;
 
 		if (
 			limit !== undefined
 			&& limit >= 0
-			&& (
-				(value !== undefined ? value.length : 0) + newFiles.length + pendingWithoutError > limit
-			)
+			&& (value !== undefined ? value.length : 0) + newFiles.length > limit
 		) {
 			setOpenSnackbar(true);
 			return;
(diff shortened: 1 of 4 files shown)


src/forms/FormFieldUpload/FormFieldUpload.tsx
```diff
@@ -200,8 +200,14 @@ const FormFieldUpload = (props: MosaicFieldProps<"upload", UploadFieldInputSetti
 	};
 
 	const shouldDisableField = useMemo(() => {
-		return fieldDef.disabled || (limit !== undefined && value?.length >= limit)
-	}, [fieldDef.disabled, limit, value]);
+		const numOfPendingFiles = Object.keys(pendingFiles).length;
```
mauriciodeleonc, line 203: The pending files object also includes those that have error'd, this means the field should get disabled **only** when the files are loading and only take into account those that don't have error messages. For example, let's assume the file limit is 3, the user uploads 3 files, while they're loading the filed is disabled, then all finish loading but 1 throws an error message, with the code above the field would still be disabled.
 ```suggestion
		const numOfPendingFiles = Object.values(pendingFiles).filter((pendingFile: {error: string}) => pendingFile.error === undefined).length;
```

File: src/forms/FormFieldUpload/FormFieldUpload.tsx [hunk 2]
```diff
@@ -200,8 +200,14 @@ const FormFieldUpload = (props: MosaicFieldProps<"upload", UploadFieldInputSetti
...
+
+		return (
+			fieldDef.disabled ||
+			(limit !== undefined && value?.length >= limit ||
```
mauriciodeleonc, line 207: I think there's some repeated code here that could be collapsed. If we think of these "or" statements as venn diagrams we have 3 intersects, which could be combined into 1 single operation.
```
return (
			fieldDef.disabled ||
			(
				limit !== undefined
				&& limit >= 0
				&& (value !== undefined ? value.length : 0) + numOfPendingFiles >= limit
			)
		)
```

pendingFiles will always be an object so it's length will always be a number, the only thing we'd need to check for would be if value exists, and if it does, we put it's length, if not then it's a 0

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #381, opened 2023-02-22
MOS-963: Tooltip z-index is incorrect

cristianvj23:
# What's included?
- Updated the button tooltip to utilize z-index: 1500


mauriciodeleonc (changes requested):
During refinement we mentioned we were going to add a tooltip to the "Open new form" button in Drawers because then there wouldn't be a way of testing if this actually works. The tooltip could say something like "Clicking opens form in a new drawer".

PR #388, opened 2023-02-24
MOS-969: Icon Card type updated 1

JAndresHJGlobant:
## What's included?

- titleIcon Card's prop change from JSX.Element to SvgIconComponent


mauriciodeleonc (changes requested):
Tests are failing because they're still expecting a JSX.Element

mauriciodeleonc (approved):

PR #402, opened 2023-03-07
MOS-992: Phone field retuns invalid data

JAndresHJGlobant:
## What's included?

Fixes phone dropdown field to save undefined in the Form value when the only number in the field is the dial code.

### DIFF
--- a/src/forms/FormFieldPhoneSelectionDropdown/FormFieldPhoneSelectionDropdown.tsx
+++ b/src/forms/FormFieldPhoneSelectionDropdown/FormFieldPhoneSelectionDropdown.tsx
@@ -27,11 +27,11 @@ const FormFieldPhoneSelectionDropdown = (
 	const [dialCode, setDialCode] = useState("");
 
 	const onPhoneChange = (phoneValue: string, data: CountryData) => {
-		onChange(phoneValue);
-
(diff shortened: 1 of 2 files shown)


src/forms/FormFieldPhoneSelectionDropdown/FormFieldPhoneSelectionDropdown.tsx
```diff
@@ -24,6 +24,17 @@ const FormFieldPhoneSelectionDropdown = (
 		value,
 	} = props;
 
+	const [dialCode, setDialCode] = useState("");
+
+	const onPhoneChange = (phoneValue: string, data: CountryData) => {
+		onChange(phoneValue);
```
mauriciodeleonc, line 30: This approach executes the onChange twice when the user deletes all the text, but it's too quick to be noticeable. If you use breakpoints you'll see it goes to the fieldOnChange action twice, which we could avoid to prevent sending data twice.

I suggest either moving the if statement to the top of the function and then return, or do an else statement.

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #403, opened 2023-03-08
MOS-989: DefaultValue not accepting false as value

cristianvj23:
# What's included?
- Updated form component to accept false as a value


mauriciodeleonc (changes requested):
All the errors and warnings that the FormFields threw are expected, so the only file that should've changed is Form.tsx. For example, with the FormFieldText it now says that it's possible to pass "false" as a value, which doesn't make much sense because we know that's impossible as we declared it to only receive strings (or undefined when not passed).
The reason it was throwing all those errors was because of how the defaultValues were declared in Form.stories.tsx. They were declared using the "x && y" which basically says "if x, then return y, else return false", so all fields were assigning false to their defaultValues. In a real life scenario this doesn't make much sense because we know not all fields receive booleans as value, and we have typescript in place to catch those errors in case they manually decide to pass a boolean, like this:
image

mauriciodeleonc (approved):

PR #405, opened 2023-03-09
MOS-987: Add tooltips to Chips

cristianvj23:
# What's include?
- Added a tooltip to the deletable chip component when its width exceeds 186px.

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { ReactElement, HTMLAttributes } from "react";
+import { ReactElement, HTMLAttributes, useRef, useState, useEffect } from "react";
 
 // Material UI
 import ClearIcon from "@mui/icons-material/Clear";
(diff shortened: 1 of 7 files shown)


src/components/Chip/Chip.tsx
```diff
@@ -11,17 +12,28 @@ import { StyledChip, StyledDeletableChip, StyledLabel } from "./Chip.styled";
...
-			label={<StyledLabel>{label}</StyledLabel>}
-			{...props}
-			required={required}
-			disabled={disabled}
-			selected={selected}
-			deleteIcon={<ClearIcon data-testid='delete-icon-test-id' />}
-			onDelete={onDelete}
-			data-testid="delete-chip-testid"
-		/>
+		<Tooltip text={deletableChipWidth >= 186 ? label : ""}>
```
mauriciodeleonc: This works only for fixed-width chips (for example in chip playground / kitchen sink) but doesn't for the advanced selection once you start adding more because then the scrollbar appears and makes the chips get smaller by a few pixels making the text still show the "..." but not reaching the 186 px. I feel we could just keep the tooltip to always render when using deletable chips.

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #406, opened 2023-03-09
MOS-981: Address field is losing the typed in number

cristianvj23:
# What's include?
- Added street_number to address field


src/forms/FormFieldAddress/AddressDrawer/AddressDrawer.tsx
```diff
@@ -293,8 +294,7 @@ const AddressDrawer = (props: AddressDrawerProps): ReactElement => {
 			console.warn('Country response from google "' + addressComponentsMap.country.label + '" could not be found in the list of countries provided in getOptionsCountries.');
 			componentsNotFound += `${componentsToAddress.country}, ${componentsToAddress.administrative_area_level_1}, `;
 		}
-
-		await setFieldValue("address1", addressComponentsMap.route.label, true);
+		await setFieldValue("address1", `${addressComponentsMap.street_number.label} ${addressComponentsMap.route.label}`, true);
```
mauriciodeleonc: This assumes that street_number will always be present, so whenever the user adds an address that doesn't have a number it results in an empty space 
image

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #421, opened 2023-03-23
Mos 1010 Update disabled state for text like fields

mauriciodeleonc:
# What's included?
- Extracted StyledDisabledText component to be reused by text-like fields
- Updated FormFieldText, FormFieldPhoneSelectionDropdown, and FormFieldDate to s
- Updated DataViewFilterText, DataViewFilterTextDropdownContent, and FormFieldNumberTable to work with disabled styling changes.


JAndresHJGlobant (approved):
Nice work, I just want to point out the following:

- When working on the QA/Automation process the field label will still gray out when it is disabled and that will be solved when https://github.com/simpleviewinc/sv-mosaic/pull/419 be merged.

- We can safely close https://simpleviewtools.atlassian.net/browse/MOS-1013 since as we had anticipated the update of the disabled state of the text field will automatically fix the NumberTable.

PR #427, opened 2023-04-05
MOS-1014: Update disabled state for card components

JAndresHJGlobant:
## What's included?

Updates the disabled state for the following fields:

- Address
- ImageVideoDocument
- MapCoordinates
- ImageUpload


mauriciodeleonc (changes requested):
The following tests showed errors
image

image
image
image
image
image
image
image
image

PR #428, opened 2023-04-11
MOS-1011: Update disabled state for selection fields

JAndresHJGlobant:
## What's included?

Updates the disabled state of the following form fields:

- Checkbox
- ChipSingleSelect
- Radio
- DropdownSingleSelect
- AdvancedSelection


mauriciodeleonc (changes requested):
I believe we should make the CheckboxList receive and return the same type. Right now it receives string[] but returns MosaicLabelValue[], they both should be MosaicLabelValue[].

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #434, opened 2023-04-18
MOS-1022 Validate correct height between regular and multiline text fields

mauriciodeleonc:
# What's included?
- Cleaned code in FormFieldText.styled.tsx to align heights between the regular and multiline variants.

### DIFF
--- a/src/forms/FormFieldText/FormFieldText.styled.tsx
+++ b/src/forms/FormFieldText/FormFieldText.styled.tsx
@@ -10,70 +10,70 @@ import theme, { Sizes } from "@root/theme";
 export const StyledTextField = styled(({ fieldSize, ...rest }) => (
 	<TextField {...rest} />
 ))`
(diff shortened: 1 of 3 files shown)


src/forms/FormFieldText/FormFieldText.styled.tsx
```diff
@@ -26,14 +26,16 @@ export const StyledTextField = styled(({ fieldSize, ...rest }) => (
 	}
   }
 
-  .MuiOutlinedInput-multiline, input.MuiOutlinedInput-input {
-    padding: ${theme.fieldSpecs.inputText.padding};
+  .MuiInputBase-multiline {
+	padding: 0px;
```
JAndresHJGlobant, line 30: I just noticed that in some places the code is spaced using tabulations and in others uses spaces. I think we should use tabs for all the file since that's our linter configuration and when we run it we should not have any issues.

JAndresHJGlobant (changes requested):

JAndresHJGlobant (approved):

PR #438, opened 2023-04-21
MOS-1024: Add show capability to content buttons

JAndresHJGlobant:
## What's included?

- Adds show capability to the buttons passed to the Content component
- Adds unit test for the `showContent` function
- Fixes issue with the `showContent` that was not being applied when a single `false` value was passed, now is explicitly only checking for `undefined` values

### DIFF
--- a/src/components/Content/Content.tsx
+++ b/src/components/Content/Content.tsx
@@ -25,7 +25,7 @@ import { ActionAdditional } from "../DataView";
  * @param show is used to evaluate whether a field should be displayed
  * @returns boolean flag comming from the  validation of the show paramater.
  */
-const showContent = (show: ActionAdditional["show"]) => {
+export const showContent = (show: ActionAdditional["show"]) => {
 	const conditions = isArray(show) ? show : [show];
 
 	const isValid = conditions.every((show) => {
@@ -74,7 +74,7 @@ const Content = (props: ContentProps): ReactElement => {
 			);
 		}
 
-		if (currentField?.show && !showContent(currentField?.show)) {
+		if (currentField?.show !== undefined && !showContent(currentField?.show)) {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


src/components/Content/Content.tsx
```diff
@@ -144,12 +144,17 @@ const Content = (props: ContentProps): ReactElement => {
...
-							<Button key={`${button.label}-${idx}`} {...button} />
-						))}
-					</ButtonsWrapper>}
+				{buttons && (
+					<ButtonsWrapper
+						className={cardVariant ? "card-buttons" : "standard-buttons"}
+					>
+						{buttons?.map((button, idx) =>
+							button.show !== undefined ? (
+								showContent(button.show) && <Button key={`${button.label}-${idx}`} {...button} />
```
mauriciodeleonc, line 153: I think we might want to memoize this outside of the render fn and just render the valid buttons, that way we don't have to run all show functions with every change.
Something like:
`const validButtons = useMemo(() => {

}, [buttons]);`

mauriciodeleonc (changes requested):

src/components/Content/Content.tsx
JAndresHJGlobant: As we were talking using arrays or objects as dependencies is not possible because of the way they are compared, there has been a proposition to be able to use a custom comparator function that would be used to determine whether the dependencies of the hook have changed, instead of the default shallow equality check. https://github.com/facebook/react/issues/14476

I made a record with the approach proposed and you can see that the useMemo computation is being made with every re-render.
memoizing-references

So I would suggest not trying to memoize this computation since the `useMemo` already has a small overhead when it checks if its dependencies changed. Moreover, I'd argue that this computation is not very expensive, and it's very unlikely that tons of buttons want to be rendered.

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #440, opened 2023-04-21
TAE - MOS-1049 - Standardize automation tests

torres-nahuel:
### Jira Ticket
https://simpleviewtools.atlassian.net/browse/MOS-1049

### What's included?
Due to internal changes in the project, the automated tests have different naming conventions, some are inside folders, some do not, some use camel casing, some snake casing, etc.

The objective of the this PR is to make them all match the same naming conventions.


mauriciodeleonc (changes requested):
All tests seem to be passing expect 1:
image
This could be one of those tests where once you run it again it passes but I'll add it just as an extra check.

I also have another comment. Some tests still have differences in their casing, for example in the picture you'll see that the Content folder starts with upper case, but dataview starts with lower case, also inside some folders, some tests start with upper and others with lower.

image

PR #441, opened 2023-04-21
Mos 1005 - Map out correct types for DataView sub components

mauriciodeleonc:
# What's included?
- Moved all DV subcomponents to their own folders. Each folder includes (as minimum) the component file (_component_.tsx), it's types file, and an index. Doing this also means all files importing these components have been updated.
- Moved some types / interfaces from DataViewTypes to their corresponding component. For example if DVTypes had a type called "DataViewTrDef" that was only used in DataViewTr, then it was moved to that component's types file.
-  Removed unused props from several types / interfaces. E.g. DataViewFilterMultiselectDropwodnContent had a "value" prop, but the component didn't use that prop anywhere, so there was no reason for it to be expected in the interface.
- Cleaned props. Previously all props used to be optional (most of them will still be due to the nature of the DataView, as it only really requires few props), but now components have been updated to have optional and required props where needed.


src/components/DataView/DataViewActionsRow/DataViewActionsRow.tsx
```diff
@@ -7,7 +7,7 @@ import DataViewPager from "../DataViewPager";
 import { DataViewActionsRowWrapper, LeftControlsContainer, RightControlsContainer } from "./DataViewActionsRow.styled";
 import { DataViewActionsRowProps } from "./DataViewActionsRowTypes";
 import Checkbox from "../../Checkbox";
-import DataViewBulkActionsButtonsRow from "../DataViewBulkActionsButtonsRow";
+import DataViewBulkActionsButtonsRow from "../DataViewBulkActionsButtonsRow/DataViewBulkActionsButtonsRow";
```
JAndresHJGlobant, line 10: I think this import can be done:
`import DataViewBulkActionsButtonsRow from "../DataViewBulkActionsButtonsRow";`
Since the `DataViewBulkActionsButtonsRow` has an index file.

src/components/DataView/DataViewActionsRow/DataViewActionsRowTypes.ts
```diff
@@ -1,13 +1,13 @@
...
-	DataViewDisplay,
 	DataViewOnSortChange,
 	DataViewProps,
 } from "../DataViewTypes";
+import { DataViewDisplay } from "../DataViewDisplays";
 export interface DataViewActionsRowProps {
 	bulkActions?: DataViewProps["bulkActions"];
-	checked?: boolean[];
+	checked?: DataViewProps["checked"];
 	display: string;
```
JAndresHJGlobant, line 11: The display prop can also be referenced like `DataViewProps["display"]`

src/components/DataView/DataViewBulkActionsButtonsRow/DataViewBulkActionsButtonsRowTypes.ts
```diff
@@ -0,0 +1,9 @@
+import { MosaicObject } from "@root/types"
+import { DataViewBulkAction } from "../DataViewTypes"
+
+export interface DataViewBulkActionsButtonsRowProps {
+	bulkActions: DataViewBulkAction[]
+	data: MosaicObject[]
+	checked: boolean[]
+	checkedAllPages: boolean
```
JAndresHJGlobant, line 8: If the props  `checked` and `checkedAllPages` come from the point where the DV is instantiated I think we can reference them using its interface. `DataViewProps["checkedAllPages"];` and `DataViewProps["checked"];`

src/components/DataView/DataViewColumDrawer/DataViewColumnDrawerTypes.ts
```diff
@@ -1,16 +1,12 @@
 import { DataViewProps } from "../DataViewTypes";
 
-export interface DataViewColumnDrawerProps {
-	open?: boolean;
-	onClose?: () => void;
-	columns?: DataViewProps["columns"];
-	allColumns?: DataViewProps["columns"];
-	onChange?: DataViewProps["onColumnsChange"];
+export interface DataViewColumnDrawerProps extends DataViewColumnDrawerContentProps {
+	open: boolean;
```
JAndresHJGlobant, line 4: It looks like we have extra spaces here.

src/components/DataView/DataViewTHead/DataViewTHead.tsx
```diff
@@ -1,17 +1,16 @@
...
 import styled from "styled-components";
-// import jsvalidator from "jsvalidator";
 
 import SwitchLeftIcon from "@mui/icons-material/SwitchLeft";
 import SwitchRightIcon from "@mui/icons-material/SwitchRight";
 
 import Checkbox from "@root/components/Checkbox";
-import DataViewBulkActionsButtonsRow from "../DataView/DataViewBulkActionsButtonsRow";
-import DataViewBulkAllBar from "../DataView/DataViewBulkAllBar";
+import DataViewBulkActionsButtonsRow from "../DataViewBulkActionsButtonsRow/DataViewBulkActionsButtonsRow";
```
JAndresHJGlobant, line 8: This import can also be reduced to `import DataViewBulkActionsButtonsRow from "../DataViewBulkActionsButtonsRow";`

src/components/DataViewFilterMultiselect/DataViewFilterMultiselect.tsx
```diff
@@ -115,7 +115,6 @@ function DataViewFilterMultiselect(props: DataViewFilterMultiselectProps) {
 				onExited={onExited}
 			>
 				<DataViewFilterMultiselectDropdownContent
-					value={value}
```
JAndresHJGlobant, line 118: Just curious, was this prop not being used by the `DataViewFilterMultiselectDropdownContent`?

JAndresHJGlobant (changes requested):

PR #443, opened 2023-05-02
MOS-1044 DataView activeFilters displaying in the wrong order

mauriciodeleonc:
# What's included?
- Updated logic for rendering activeFilters in DataViewFilters.tsx
- Added knob for showing active filters loaded on a different order than the regular filters array.

### DIFF
--- a/src/components/DataView/DataViewFilters/DataViewFilters.test.tsx
+++ b/src/components/DataView/DataViewFilters/DataViewFilters.test.tsx
@@ -1,0 +1,105 @@
+import React from "react";
+import { cleanup, render, screen } from "@testing-library/react";
+import DataViewFilters from "./DataViewFilters";
+import DataViewFilterText from "@root/components/DataViewFilterText";
+import DataViewFilterMultiselect from "@root/components/DataViewFilterMultiselect";
+import DataViewFilterDate from "@root/components/DataViewFilterDate";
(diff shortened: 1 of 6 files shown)


JAndresHJGlobant (approved):

src/components/DataView/DataViewFilters/DataViewFilters.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+		for (let i = 0; i < buttons.length; i++) {
+			if (i === updatedIndex)
+				realOrder.push(buttons[updatedIndex].textContent.toLowerCase());
+			if (i === titleIndex)
+				realOrder.push(buttons[titleIndex].textContent.toLowerCase());
+			if (i === keywordIndex)
+				realOrder.push(buttons[keywordIndex].textContent.toLowerCase());
+		}
+
+		expect(JSON.stringify(realOrder)).toEqual(JSON.stringify(activeFilters));
```
JAndresHJGlobant, line 103: I think we can safely remove the application of `JSON.stringify` to both arrays the `realOrder `and the `activeFilters `and the test should be good to go. Having something like: 

`expect(realOrder).toEqual(activeFilters);`

I tried that on my local and the test still passes.

JAndresHJGlobant (changes requested):

JAndresHJGlobant (approved):

PR #445, opened 2023-05-03
MOS-1040: Update NumberTable to support intl.Number format options

JAndresHJGlobant:
## What's included?

Implements support for intl.Number format options in the NumberTable component.

### DIFF
--- a/automation_testing/tests/FormFields/FormFieldNumberTable/FormFieldNumberTable.spec.ts
+++ b/automation_testing/tests/FormFields/FormFieldNumberTable/FormFieldNumberTable.spec.ts
@@ -72,7 +72,7 @@ test.describe.parallel("FormFields - FormFieldNumberTable - Playground", () => {
 		expect(await ffNumberTablePage.numberTableLocator.locator("thead").locator("th").last().textContent()).not.toBe("No. Rooms");
 	});
 
-	test("Validate that the different ", async () => {
+	test("Validate that the different format are displayed correctly.", async () => {
 		const formatsToValidate = ["USD", "EUR", "JPY", "GBP"];
 		for (let i = 0; i < formatsToValidate.length; i++) {
 			await ffNumberTablePage.visitPageWithNumberFormat(formatsToValidate[i]);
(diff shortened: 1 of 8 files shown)


src/forms/FormFieldNumberTable/FormFieldNumberTable.stories.tsx
```diff
@@ -26,6 +26,11 @@ export const Playground = (): ReactElement => {
 	const helperText = text("Helper text", "");
 	const displayColumnsSums = boolean("Display columns sums", true);
 	const displayRowsSums = boolean("Display rows sums", true);
+	const currency = select(
```
mauriciodeleonc, line 29: I would suggest changing the name of this const / knob to "formatOptions", that way if anyone reads the code in storybook they'll understand it a bit better.
```suggestion
	const formatOptions = select("Number Format Options", ["{ style: 'currency', 'USD' }", "{ style: 'currency', 'EUR' }", ...], "{ style: 'currency', 'USD' }");
```

With this approach devs will understand better the whole object they would need to pass and it makes it easier for you in line 52 to just pass the variable

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

automation_testing/tests/FormFields/FormFieldNumberTable/FormFieldNumberTable.spec.ts
```diff
@@ -72,6 +72,15 @@ test.describe.parallel("FormFields - FormFieldNumberTable - Playground", () => {
 		expect(await ffNumberTablePage.numberTableLocator.locator("thead").locator("th").last().textContent()).not.toBe("No. Rooms");
```
mauriciodeleonc, line 72: Got several errors in this file. It's expecting Not a Number, which I assume might be happening because it's reading the numbers as prices (e.g. "$123.00") and then tries to add them with the "$" which causes it to break. I think you might be able to solve this using something like 
inputSum += Number(await this.tableBodyRowLocator.nth(j).locator("input").nth(i).inputValue()**_.split("$")[1]_** );

image
image
image
image
image

mauriciodeleonc (changes requested):

mauriciodeleonc (approved):

PR #447, opened 2023-05-05
MOS-1043 Allow form sections to be collapsed

mauriciodeleonc:
# What's included?
- Updated "FormLayout" and "Playground" stories to make it easier for QA engineers to test how sections can open / close when a field has an error.
- Added "collapsed" prop to the Section component.
- Updated Section component to use MUI's Accordion component.
- Added logic to expand collapsed Sections when they have an error'd field inside.
- Updated formAction in charge of scrolling to fields with errors to wait until the section has expanded.


JAndresHJGlobant (changes requested):
We should also add the respective documentation about the new Form `collapsed` prop in the `Form.stories.mdx` file. By taking a look at our docs we only have a reference to the `SectionDef` sections type but nowhere are we specifying what this type contains, maybe we could add a new subsection after the `Dispatch` section to describe `SectionDef` in detail.

JAndresHJGlobant (approved):

PR #449, opened 2023-05-09
MOS-1006: Fixes AdvancedSelection and DVMultiSelect components' padding

JAndresHJGlobant:
## What's included?

Fixes side paddings on DVFilterMultiselect and AdvancedSelection


mauriciodeleonc (changes requested):
All changes lgtm but found a visual bug that causes scrollbars to appear when a tooltip with long text renders on the screen because it pushes the available space.
image

mauriciodeleonc (approved):

mauriciodeleonc (approved):

PR #454, opened 2023-05-18
MOS-1057 Add data-testid to Grid columns

mauriciodeleonc:
# What's included?
- Modified DataViewTd's types to accept aria-labels
- Updated DataViewTr to pass aria-label to Td's
- Updated DataViewTd to use aria-labels


mauriciodeleonc (commented):
I'm getting the following error in Filter.spec.ts 
image

PR #463, opened 2023-06-21
MOS-1056 Show capability

GeorgeBatt-SV:
- Factor out `filterActions` function to a more generic "show" evaluator
- Consume show evaluator with components that previous used `filterActions`: Content, DataView actions 
- Consume show evaluator with fields, introducing support for conditional fields


src/components/Content/Content.test.tsx
```diff
@@ -221,18 +222,18 @@ describe("Content componenent with no buttons", () => {
 
 describe("showContent helper function", () => {
```
Porl91, line 223: When I was doing my training with Mike & Colin were keen on exhaustive tests (i.e. any test in a function would normally have a separate test). It's probably a bit ambitious retro-fitting 100% code coverage, but for these sorts small utility functions it probably make sense:

image

Porl91: Reckon it makes sense to merge these parameterless tests into `src/utils/show/evaluateShow.test.ts` just so they're in one place?

GeorgeBatt-SV: Introduced exhaustive tests and isolated them to the evaluateShow test suite. Review re-requested.

Porl91 (approved):

PR #464, opened 2023-06-22
MOS-1086 Introduced and implemented a generic Blank component

GeorgeBatt-SV:
This PR introduces a new "Blank" component. This component will be used across all components designed to display a value that can be undefined, an empty string, or any other value considered "empty". It is being consumed by the Field Col and Content components.

### DIFF
--- a/src/components/Content/Content.tsx
+++ b/src/components/Content/Content.tsx
@@ -20,6 +20,7 @@ import {
 } from "./Content.styled";
 import Button from "@root/components/Button";
 import { ActionAdditional } from "../DataView";
+import Blank from "@root/components/Blank";
 
 /**
  * @param show is used to evaluate whether a field should be displayed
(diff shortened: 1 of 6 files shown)


Porl91 (changes requested):

src/components/Content/Content.tsx
```diff
@@ -79,6 +80,15 @@ const Content = (props: ContentProps): ReactElement => {
 		}
 
 		const fieldName = currentField?.column ? currentField?.column : currentField?.name;
+		let fieldValue = data[fieldName];
```
Porl91, line 83: Feel free to resolve this comment because you haven't changed this part but you could eliminate those optional chainings here because we've already done the existence check above.

We could also move the `if (!field) {` check to the top of the method to avoid the `.find()` call & bail out as early as possible.

image

GeorgeBatt-SV: Good call. I'm looking to refactor this component for MOS-1098, so I'll address this change when working on that ticket.

Porl91 (approved):

PR #465, opened 2023-06-22
MOS-1058 bug scrub

GeorgeBatt-SV:
This PR

- Removes the state dependency within the Advanced Selection form field component
- Removes behaviour that prompts user about unsaved changes when closing the draw
- Prevents clicks to the draw backdrop from closing the draw

### DIFF
--- a/src/components/DataViewFilterMultiselect/DataViewFilterMultiselectDropdownContent.tsx
+++ b/src/components/DataViewFilterMultiselect/DataViewFilterMultiselectDropdownContent.tsx
@@ -33,20 +33,13 @@ function DataViewFilterMultiselectDropdownContent(props: DataViewFilterMultisele
 
 	const [showCreateOptionButton, setShowCreateOptionButton] = useState(false);
 
-	const [disabled, setDisabled] = useState(false);
-
 	const { t } = useMosaicTranslation();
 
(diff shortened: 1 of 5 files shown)


GeorgeBatt-SV:
@Porl91 In some cases, the `DataViewFilterMultiselectDropdownContent` component is used without the integrated Apply/Clear buttons, so I think this is why the component emits the changes up the tree.

It is demonstrated in [`AdvancedSelectionDrawer`](https://github.com/GeorgeBatt-SV/sv-mosaic/blob/MOS-1058-bug-scrub/src/forms/FormFieldAdvancedSelection/AdvancedSelectionDrawer.tsx#L77): `DataViewFilterMultiselectDropdownContent` is used with the `hideButtons={true}` property, and the save action becomes the `PageHeader`s responsibility instead. For that to work, the `AdvancedSelectionDrawer` needs an up-to-date list of the selected options.

Porl91:
@GeorgeBatt-SV Ah gotcha, reckon it makes sense to avoid calling `props.onChange` in `onChange` / `handleToggle` if `hideButtons` is false?

I'm just wondering if consumers that have the apply button showing would expect the selection to only be propagated when you user clicks 'Apply'.

Porl91 (approved):

PR #466, opened 2023-06-23
MOS-1089 textfield height

GeorgeBatt-SV:
This PR removes the padding, line height and height styling from the FormFieldText

### DIFF
--- a/src/forms/FormFieldText/FormFieldText.styled.tsx
+++ b/src/forms/FormFieldText/FormFieldText.styled.tsx
@@ -26,26 +26,16 @@ export const StyledTextField = styled(({ fieldSize, ...rest }) => (
 		}
 	}
 
-	.MuiInputBase-multiline {
-		padding: 0px;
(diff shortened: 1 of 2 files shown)


src/forms/FormFieldText/FormFieldText.styled.tsx
```diff
@@ -26,16 +26,10 @@ export const StyledTextField = styled(({ fieldSize, ...rest }) => (
 		}
 	}
 
-	.MuiInputBase-multiline {
-		padding: 0px;
-	}
-
 	.MuiOutlinedInput-input {
```
Porl91, line 29: I'm checking checking how this comes through in the Form playground and it looks like the height & padding have increased post-changes. Do you reckon we should constrain these back their originals for single-line fields just so  the heights/paddings match up with other components like dropdowns/datepickers?
image

Porl91 (approved):

PR #468, opened 2023-06-26
MOS-757 CircleCI

GeorgeBatt-SV:
CircleCI experiments


owenallenaz:
Alright, reviewed this initial operation and I have some thoughts. I think you are on a decent path here just some things to be aware of:

* The reason for the prior workflows section of the config was for fan-out to parallelize some of the steps. This allows us to build and then perform steps like storybook, publish in parallel. You may or may not want to keep this mechanism. The need to split steps into their own steps definitely adds complexity. For example, if we just, didn't do that... then this whole thing could be dramatically simplified, even using the prior config because the same container that will publish will also build. In this case the slowest part of the operation by far is the unit tests and the build so I don't know if the parallelization changes much in practicality. Another factor here is also the type of unit tests. For example the unit tests that depend on the playwright system can parallelize with the embedded RTL unit tests. If you were to go to one singular executor then you can't parallelize those steps well. So given that... I think you'll still want parallelization.
* The intent behind docker vs executor was simple to ensure that the local and cloud environment is as close to equal as possible. Using docker, we know they are nearly identical since the same Dockerfile builds them both. For example, if you look at your current code, in the Dockerfile, we're installing jq and gettext-base and a specific version of Node and yarn. This means if you update the node version in one place, you also need to update it in the circleci file. Same with yarn and other dependencies. If you don't use the docker container you have to be diligent about ensuring that both steps are kept in sync. It's a trade-off. My suspicion when I looked at this in the beginning was that since you're already having to do a shared workspace, then the addition of docker, on that shared workspace shouldn't be much more since I believe it's a single command to dump the built docker image to the workspace, and a single command to run it, but I didn't actually attempt it myself. So I'd say that the Docker variant is **preferred**, if you can get it working, just so that you can ensure that local and cloud are as close to identical as possible. But, if the choice is units that don't run, and using executor instead of `docker` then working is better than not working. So maybe give it a shot, and if it spirals on you massively abandon ship.
* Mau mentioned that he was able to run a circleci command locally, that allows you interate without having to push code to github, wait for the trigger. In that scenario you might be able to more easily debug specific parts of the config, because you could, for example, on your local box alter the unit tests and build commands so that they don't do anything and finish in 0ms, this way you can ensure the orchestration of your circleci config is correct. This is really tricky if you're having to git push everytime and that's what resulted in my pile of configs when I was initially setting this up. I've personally never tried running circleci locally (or inside the sv-kube VM), but it's worth exploring.

PR #470, opened 2023-06-30
MOS-1056 Add show behaviour to sections

GeorgeBatt-SV:
This PR is a follow on from https://github.com/simpleviewinc/sv-mosaic/pull/463 and brings the _show_ behaviour to form sections as well as fields.

### DIFF
--- a/src/components/Field/RegisteredField.tsx
+++ b/src/components/Field/RegisteredField.tsx
@@ -1,0 +1,20 @@
+import React, { ReactElement, memo, useEffect } from "react";
+
+import { MosaicFieldProps } from ".";
+import Field from "./Field";
+import { formActions } from "../Form";
+
+const RegisteredField = ({
+	dispatch,
+	name,
+	...props
+}: MosaicFieldProps<any> & {dispatch: any, name: string}): ReactElement => {
+	useEffect(() => {
+		dispatch(formActions.mountField({name}));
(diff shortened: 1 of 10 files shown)


GeorgeBatt-SV:
@Porl91 Good call - it's not just your setup. I've checked in a change that filters sections that are provided to the `TopComponent` so that nav items don't appear for sections that aren't in the DOM.

It's not particularly pretty, and it still has scroll-based bugs (which existed before the introduction of this show behaviour). These bugs and the scroll behaviour as a whole are already being addressed in [another ticket](https://simpleviewtools.atlassian.net/jira/software/c/projects/MOS/boards/149?modal=detail&selectedIssue=MOS-975).

This is just to prevent errors being thrown on this branch.

Porl91:
Yalright @GeorgeBatt-SV Do you get this error when you load the "Conditional Sections" page?

image

It looks like that `sections` variable is `false` for me so the `.filter()` call gives an error:

image

GeorgeBatt-SV:
@Porl91 As discussed in Slack, this was due to (non-strict) TS not detecting an incorrect type being passed to the `sections` property in the Form playground story.

I've updated that story to have that parameter fallback to `undefined` if `showSections` is not more than `0` (where it would have been `false` previously)

Porl91 (approved):

Porl91:
LGTM!

PR #471, opened 2023-07-04
MOS-1090 ButtonRow modifications and usage

GeorgeBatt-SV:
This PR slightly modifies the ButtonRow component to allow it to be used across more components where a list of buttons is possible, improving consistency. It adds support for button separators - in this case a simple grey pipe - to make it diverse enough to reuse where the context calls for it.

### DIFF
--- a/src/components/ButtonRow/ButtonRow.styled.tsx
+++ b/src/components/ButtonRow/ButtonRow.styled.tsx
@@ -11,7 +11,11 @@ const gapMap: Record<ButtonRowProps["gap"], string> = {
 export const Row = styled.div`
 	display: flex;
 	align-items: center;
-	flex-wrap: wrap;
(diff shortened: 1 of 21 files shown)


Porl91:
@GeorgeBatt-SV  It looks like the action buttons in the `DataView` have become a bit wonky with this change:
image

Previously it looked like they lined up:
image

Was there any additional styling we were meant to add for this page to keep those buttons inline?

Page: http://kube.simpleview.io:10000/?path=/docs/components-dataview--playground

GeorgeBatt-SV:
Cheers @Porl91, removed the default wrapping and added a `wrap` prop to the `ButtonRow` for places where it is needed instead.

Porl91 (approved):

PR #491, opened 2023-07-31
MOS-1120 Implement a form loading queue

GeorgeBatt-SV:
This PR adds support for a form loading queue. This can be used to prevent submissions when one or more components is in a loading state. It also has the `FormFieldUpload` component utilise the new loading queue functionality to prevent users from submitting the form whilst file uploads are in progress.

### DIFF
--- a/src/components/Form/formActions.ts
+++ b/src/components/Form/formActions.ts
@@ -199,16 +199,16 @@ export const formActions = {
 	},
 	submitForm() {
 		return async function (dispatch, getState, extraArgs): Promise<{ valid: boolean; data: any; }> {
-			const { disabled, data, mounted, loading } = getState();
+			const { disabled, data, mounted, busy } = getState();
 
 			if (disabled)
 				return;
 
-			const loadingMessages = Object.values(loading).filter(Boolean);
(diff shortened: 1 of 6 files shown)


Porl91:
@GeorgeBatt-SV It looks like I get an error on this line when I submit when there aren't any `busyFields` (i.e. `undefined`):
image

Reckon we should bail out early in this case?

src/components/Form/formActions.ts
```diff
@@ -263,6 +292,38 @@ export const formActions = {
 			})
 		}
 	},
+	isSubmittable () {
+		return function (dispatch, getState): boolean {
```
Porl91, line 296: #

Porl91 (changes requested):

src/components/Form/formActions.ts
GeorgeBatt-SV: I've checked in a change that ensures the form's initial state contains `busyFields` as an empty object - it was previously `loading` but it was decided that `busyFields` is more appropriate.

Porl91 (approved):

PR #493, opened 2023-08-01
MOS-1121 Trims FormFieldAddress address parts of whitespace

GeorgeBatt-SV:


src/forms/FormFieldAddress/AddressDrawer/AddressDrawer.tsx
```diff
@@ -206,10 +206,10 @@ const AddressDrawer = (props: AddressDrawerProps): ReactElement => {
 
 		listOfAddresses.push({
 			id: id,
-			address1: state?.data?.address1,
-			address2: state?.data?.address2,
-			address3: state?.data?.address3,
-			city: state?.data?.city,
+			address1: state?.data?.address1.trim(),
```
Porl91, line 209: Reckon we should stick another null conditional on the address line properties? I just noticed if I try and "Save" the address draw with the required fields but without interacting with the address lines 1-3 it errors on those `undefined`s:

http://kube.simpleview.io:10000/?path=/story/formfields-formfieldaddress--playground

image

Porl91 (changes requested):

src/forms/FormFieldAddress/AddressDrawer/AddressDrawer.tsx
GeorgeBatt-SV: @Porl91  Great shout. Checked in changes that employ optional chaining.

Porl91 (approved):

PR #497, opened 2023-08-07
MOS-1127 mindate support

GeorgeBatt-SV:
This PR adds support for a custom `Datepicker` minimum date. By default, MUI sets this minimum date to 01/01/1900, but setting this prop will allow for it to be earlier or later.

### DIFF
--- a/src/forms/FormFieldDate/DateField/FormFieldDate.tsx
+++ b/src/forms/FormFieldDate/DateField/FormFieldDate.tsx
@@ -145,7 +145,7 @@ const FormFieldDate = (props: MosaicFieldProps<"date", DateFieldInputSettings, D
 								type: "",
 								inputSettings: {
 									placeholder: DATE_FORMAT_FULL_PLACEHOLDER,
-									minDate: fieldDef?.inputSettings.minDate
+									minDate: fieldDef?.inputSettings?.minDate
 								},
 								required: fieldDef?.required,
 							}}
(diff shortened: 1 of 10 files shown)


Porl91 (approved):

Porl91:
@GeorgeBatt-SV Ah just noticed an error on this page
http://kube.simpleview.io:10000/?path=/story/components-form--playground

It looks like `inputSettings` is optional here so we might want another null propagation.

image


image

src/forms/FormFieldDate/DateField/FormFieldDate.tsx
```diff
@@ -144,7 +144,8 @@ const FormFieldDate = (props: MosaicFieldProps<"date", DateFieldInputSettings, D
 								label: "",
 								type: "",
 								inputSettings: {
-									placeholder: DATE_FORMAT_FULL_PLACEHOLDER
+									placeholder: DATE_FORMAT_FULL_PLACEHOLDER,
```
Porl91, line 147: !

Porl91 (changes requested):

src/forms/FormFieldDate/DateField/FormFieldDate.tsx
GeorgeBatt-SV: Cheers @Porl91, amended.

Porl91 (approved):

